PR #15, opened 2022-05-07
New module: Remove Queries.

PabloOQ:
Inspired by URLSanitizer. This module removes all queries from the URL.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/ModuleManager.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/ModuleManager.java
@@ -10,6 +10,7 @@ import com.trianguloy.urlchecker.modules.list.PatternModule;
 import com.trianguloy.urlchecker.modules.list.StatusModule;
 import com.trianguloy.urlchecker.modules.list.TextInputModule;
 import com.trianguloy.urlchecker.modules.list.VirusTotalModule;
+import com.trianguloy.urlchecker.modules.list.RemoveQueriesModule;
 import com.trianguloy.urlchecker.utilities.GenericPref;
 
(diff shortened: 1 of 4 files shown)


TrianguloY:
Thank you for taking the time to implement a new module!

At first sight everything seems fine, but give me a few days to check and test it. I will notify you here on the process (and feel free to ping me if a week passes without any activity, just in case).

TrianguloY:
Ok, I would have liked some more comments, but I see that you have basically copied/edited the ClearUrlModule.java file, and I forgot to add comments there 🤦‍♂️ so that's on me.

The rest of the code seems fine, there are some minor format issues but I can easily fix them, don't worry.

The only thing I need is to know how you want to be credited (unless you don't want). Is 'PabloOQ' ok or you prefer a different name?

I'll merge the request afterwards, and hopefully publish an update next week. Thanks again!

PabloOQ:
I accidentaly removed a check for the `authority` before the pull request, is fixed now, I also added some comments.

> The only thing I need is to know how you want to be credited (unless you don't want). Is 'PabloOQ' ok or you prefer a different name?

PabloOQ is fine. Thank you!

TrianguloY:
The new version is now in beta, will be available as stable soon!

Note: I simplified the query detection and general logic so that it worked in more urls (yours failed if there were no 'http' at the beginning for example). See the details [here](https://github.com/TrianguloY/UrlChecker/commit/5e9b39a846e16ccd3096c21f4dd0c1c18f9b8efa#diff-e96966be088f85e61e362e7c3b641c6f2018f293af85f5e28e1c7d38d47d6044R66)

Issue #9, opened 2022-02-13
New module: query parameter manager


slmb3:
Thank you for making and sharing this app, it's really nice.

My suggestion is a module that parses all the query parameters, displays them as a scrollable list, and allows you to edit or delete any query parameter from the URL from the list (by tapping or swiping or something). A delete all parameters option would be useful as well. A bonus to this feature could include ability insert new parameters as well.

Example: https://example.com/?a=1&b=2 would show a list:

* a=1
* b=2

Deleting `a=1` would transform the url to https://example.com/?b=2

slmb3:
@PabloOQ thank you! And @TrianguloY thank you for reviewing and merging! 

The feature looks great. Multiple key handling is intuitive, I like it.

What do you think about including the parameter value in the list as well?

TrianguloY:
Totally agree. I even though about displaying it decoded, and even letting it be copied or even setting it as the new url. (But didn't had time to implement it)

slmb3:
The new update looks nice! 
One more potential improvement I see is to not make the param key in all caps (foo=bar shows `REMOVE FOO bar`)

Ilithy:
> The new update looks nice! One more potential improvement I see is to not make the param key in all caps (foo=bar shows `REMOVE FOO bar`)


In the last version, the keys are well displayed in lower case, here _(version 2.6-alpha)_

![Screenshot_20220827-124622](https://user-images.githubusercontent.com/36798218/187026849-c785277b-f2cd-4d80-a813-1890773aa627.png)

TrianguloY:
The upper case buttons is a standard android feature that is enabled in most versions, but not all.

In think I can disable it.

slmb3:
Odd and interesting. Thanks!

PR #16, opened 2022-05-15
Remove Queries module update

PabloOQ:
Update to the Remove Queries module, as requested in #9. Now you can remove queries one at a time, if needed. There is an arrow next to the 'Apply' button, this arrow will expand a section with a button for each query key, tapping any button will result in removing that specific query parameter.

Things to keep in mind, if there are multiple queries with the same key, each one can be removed individually, their buttons will appear in the same order they are found.

Code for the expandable box is from `ConfigActivity.java`.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/RemoveQueriesModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/RemoveQueriesModule.java
@@ -49,7 +49,7 @@ class RemoveQueriesDialog extends AModuleDialog implements View.OnClickListener 
     private Button remove;
     private TextView more;
     private LinearLayout box;
-    private List<Button> queryButtons = new ArrayList<>();
+    private List<Button> queryKeyButtons = new ArrayList<>();
     private String[] queriesArray;
 
     private String cleared = null;
@@ -91,10 +91,10 @@ class RemoveQueriesDialog extends AModuleDialog implements View.OnClickListener 
         cleared = url.replaceAll("\\?[^#]*", "");
 
         // remove previously generated buttons
-        for (Button q : queryButtons) {
+        for (Button q : queryKeyButtons) {
             box.removeView(q);
         }
-        queryButtons = new ArrayList<>();
+        queryKeyButtons = new ArrayList<>();
 
         if (!cleared.equals(url)) {
             // query present, notify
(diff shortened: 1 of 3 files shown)


TrianguloY:
Wow, nice work!

This is a nice but bigger improvement, give me a few days to review the code. 
I'll probably give you suggestions, feel free to fix them yourself (but if you prefer just leave them and I'll make them myself)

TrianguloY:
As you probably noticed, the code was merged and I made the changes myself (sorry, I couldn't resist!).

You can see them from the latest commits, and if you see something strange or you don't like the modifications just say so! I'll be happy to discuss them.

PabloOQ:
> As you probably noticed, the code was merged and I made the changes myself (sorry, I couldn't resist!).

Yes I saw it, I did want to fix the code myself to not waste your time and to familiarize myself with the app, android and the github flow, but if it's not a problem for you then there is no need to worry.

> if you see something strange or you don't like the modifications just say so!

One thing is that the `more` gray arrow with the yellow background is now a bit hard to see.

TrianguloY:
Hmm, I put it in the label because outside was harder to notice, in my opinion.
Maybe it can be made bigger or less dim? Do you want to try? 
If you do please attach screenshots with the different variants so that they can be easily reviewed. I'll try to do the same.

PabloOQ:
I also agree that the arrow should be inside.

I found that removing the tint attribute from both the collapsed and arrows gives a better result, although because the arrow is beefier than the text it looks like it's whiter, but it isn't.

With tint:
tint

Without tint:
no_tint

From contrast-finder.org:

> [WCAG success criteria 1.4.3 ](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast)requires text to have a minimum contrast ratio of 4.5 (and 3 for large text).

[With tint:](https://app.contrast-finder.org/result.html?foreground=%23808080&background=%23A16A21&ratio=4.5&isBackgroundTested=false&algo=HSV&lang=en)
tint_contrast

[Without tint:](https://app.contrast-finder.org/result.html?foreground=%23FFFFFF&background=%23A16A21&ratio=4.5&isBackgroundTested=true&algo=HSV&lang=en)
no_tint_contrast

Also looking at it now, while a minor problem, because the arrow is bigger than the text, the background color in this module is bigger than the other ones.
Left Remove Queries, right Pattern checker:
background

TrianguloY:
The issue is that on some devices the text is not white but gray, the tint on the drawables is there to make them the same color, otherwise it looks off ... but on some devices it's the opposite (android style is such a mess...no wonder why apps use their own styles, but that increases the app size by A LOT and I don't want that).

I've tested a few different attributes, and I think the best solution is to give the text a clickable state (to look like a link).
I also tested with better backgrounds: what do you think?
image

PabloOQ:
Looks good to me.

Issue #3, opened 2022-02-08
Allow updating ClearUrls dataset


TrianguloY:
Allow fetching and downloading the latest file, save and use instead of the provided one.

PR #27, opened 2022-07-02
ClearURLs module update

PabloOQ:
As requested on #3.
This update lets the user download a custom database from the internet, by default the database URL in preferences is the one from ClearURLs, so it can be used to "update" the database.
For debugging and testing purposes [I have my own database](https://github.com/PabloOQ/Rules) with only google as provider and a new rule "testing".
There are a few things to be done, or to decide if they should be done, these are properly indicated in the code with FIXME and TODO tags.  As well as fixing and changing everything you see fit.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/ClearUrlModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/ClearUrlModule.java
@@ -24,6 +24,8 @@ import org.json.JSONException;
 import org.json.JSONObject;
 
 import java.io.BufferedReader;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -40,7 +42,7 @@ import java.util.regex.Pattern;
 
 /**
  * This module clears the url using the ClearUrl database (an asset copy)
- * TODO: allow updating the database from the official page
+ * TODO: add option to delete custom database?
  */
 public class ClearUrlModule extends AModuleData {
 
@@ -144,9 +146,9 @@ class ClearUrlConfig extends AModuleConfig {
     /**
      * Replaces the database with a new one
      */
-    private void replaceDatabase(String filename, String source, String hash, boolean checkHash, Context context){
-
-        try (FileOutputStream fos = context.openFileOutput(filename, Context.MODE_PRIVATE)){
+    private void replaceDatabase(String fileName, String source, String hash, boolean checkHash, Context context){
+        // TODO check hash
+        try (FileOutputStream fos = context.openFileOutput(fileName, Context.MODE_PRIVATE)){
             JSONObject sourceJson = readJsonFromUrl(source);
 
             fos.write(sourceJson.toString().getBytes(Charset.forName("UTF-8")));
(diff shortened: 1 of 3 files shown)


TrianguloY:
Wow, bravo.

This feature is something that I should have made from day 1, but I didn't. And I'm glad you too the time to actually implement it, thanks! 

I glanced over the code and it seems good, a bit different of how I had imagined it but if it works that's the important thing.

In any case I want to thoroughly review it, and ~~maybe~~ most probably I'll change a few things (not necessarily because they are wrong). Hopefully today I'll be able to, if not next week.
I'll let you know if I have questions about the implementation.

Thank you again 😊

TrianguloY:
Since I don't have much time to spent on the app, I'm going to publish an intermediate version with the translations (which were made a few weeks ago), this pull request and other changes will be merged for the next version, sorry. 

(aka: I'm slow but I didn't forget, I'm still on it)

TrianguloY:
OK! I reviewed the code and even though there are some possible simplifications, in general it is a good approach to the feature. However I think that from a user point of view it is maybe a bit difficult to understand.

The original feature was to update the saved database from the official repository, because the built-in file is just a copy that I manually update on each version. What I had in mind is to have just an 'update' button that will retrieve the hash, check if it's different, and if so download and save the updated file (so it is used instead).
Your solution is more configurable, and I like that, but it may be a bit confusing. I think the easier solution is to add a toggle that will show/hide the extra options, so that by default only an 'update' button and 'show advanced' checkbox are shown. It would also be nice to let the user edit manually the file as text, without needing to setup a custom server.

In any case, your solution is the perfect starting point for me to adapt and refine the feature, lets hope it takes me less than a month! 😅

(Note: I'll probably write the commits on your branch and later merge myself, don't worry about the current conflicts).

PabloOQ:
I was afraid it would be too much for the user, I never thought of hiding it, somewhat basic GUI design but it didn't cross my mind, definitely that's the way to go.
I felt that it was necessary to have the option to choose the database as it gives the power-user more options.
I considered about letting the user manually edit the database, but honestly I don't see it being comfortable. The only thing that I thought could work would be a drop-down like the one from the remove queries module when 'checking' urls, and have the option to add queries that are present or even remove queries from the database, the buttons would be highlighted red to remove and green to add. But even then, the rest of fields, like redirection or even providers can't be added/modified/removed.
Also if that were to be implemented, then everytime the database is updated it would need to check which fields have been added/removed by the user, which is probably easy if the field was added, but not so easy if it was removed. To solve this I thought of having another file which stores only what the user modifies, and then check that while iterating the database.
I didn't made any of that because I don't think a user which doesn't know how the database works would (neither should) modify it.

TrianguloY:
Ugh, no wonder I kept postponing this feature. It is HUGE (and I also added an editor, which made it worse but I just wanted to). 
I wasn't sure how to design it, and I didn't like any solution I found. In the end this is the least-ugly solution I found:
image
image image
In the module setting there are only 2 new buttons. The left one opens a dialog to configure the catalog urls to manual or automatic update (this is mostly your original work). The other one opens another dialog to displays and lets you edit the catalog, just because I wanted to.

If you have any suggestion or critics or anything to say now it's the best moment. As I've always said I'm horrible with UIs so I'm sure there is a better and far more elegant solution than this.


From a programming point of view all the new logic was moved to a new ClearUrlCatalog class, leaving the original almost untouched.
Oh, and there are 16 new strings, the translator are going to have a great time...

PabloOQ:
Great! I was thinking of adding an auto-update option too, as it should be trivial, but seems you already did that. It is also way better to have these advance features hidden, having a way to see the custom database is nice too. Definitely a big improvement over my initial implementation.

PR #55, opened 2022-08-20
Image rework

Ilithy:
* i revert (my) changes on description file (for this PR)

* Updated screenshots (in English and French version) _My goal is to make them for all available languages_

* _I cracked and added an additional badge (displaying the weight of the application), to be deleted if not desired_


TrianguloY:
Really like the new badge, I just made it link to the full github-api (that's where the size comes from)
And thanks for the screenshots update :)

Merged!

Ilithy:
Thanks, I was afraid that would be too much.
Yes, it is better to merge this one before the other I think, ready! :)

Issue #45, opened 2022-08-07
Change language in-app


TrianguloY:
Add an option to allow the user to change the language of the app by choosing from a list inside the app itself, without having to modify the system settings.

Will probably require a new app settings sub screen.

Ilithy:
> Will probably require a new app settings sub screen

_If you add this (General settings), it would be nice to have also in these settings the possibility to change the theme of URLCheck (light/dark/system)._

TrianguloY:
Yup. There are several app-related settings planned for the future, and thats exactly why I made a new settings screen (currently on a temporal branch until I have something to fill it with!).

TrianguloY:
Ah wait, I though there were already an issue for the theme selector. Feel free to create one (I'll do it later otherwise)

Ilithy:
Sorry for posting this here, this is neither good nor practical, I didn't dare to create an issue for this, it's done (#94);
I also think there was already one subject related to themes, but I did not find it either (maybe also in an issue on the peripheral subject)

PR #56, opened 2022-08-20
More screenshots

Ilithy:
* Added screenshot for languages:
  - Spanish
  - Portuguese
  - Ukrainian

_(I can't do the screenshots for Turkish language, because strangely GrapheneOS doesn't have this language pack, if someone provides me the screenshots in Turkish, I could do the downstream work, or I would wait (#45) to be able to modify the language of Url Checker directly in the application)_

If I made a mistake somewhere, let me know, thank you.


TrianguloY:
At first I was surprised about those test.txt, but then I realized that on Github web you can't edit the path of uploaded files...or can you? ;)

Yes, they can be changed! You can either just change the url yourself: https://github.com/TrianguloY/UrlChecker/upload/master/test/secret/subfolder
Or place your files from your computer in a folder and upload the folder.

The second option allows you to upload multiple files to different folders in one go (just make all the necessary subfolders) but maybe it's not supported on all browsers, the first one works even from mobile!


(Note: you don't need to change this PR, I don't mind those extra commits, just saying because probably it's easier for you when uploading files next time)

TrianguloY:
You have no idea the amount of work you just saved me. There is a way to automate making the screenshots, but I'm still working to make it work (and they don't have that nice tint you add). Now I only need to automate uploading them to Play Store, which is also a work in progress (currently I'm doing it manually).

Don't worry about the Turkish locale, as you said it will be possible when I add the ability to change the locale in-app.

Btw, I noticed that the open text in the ukranian locale touches the borders, which means that can happens on all the other locales too. I'll add a bit of padding to fix that.

Ilithy:
> At first I was surprised about those test.txt, but then I realized that on Github web you can't edit the path of uploaded files...or can you? ;)
>Yes, they can be changed! You can either just change the url yourself: https://github.com/TrianguloY/UrlChecker/upload/master/test/secret/subfolder
Or place your files from your computer in a folder and upload the folder.
>The second option allows you to upload multiple files to different folders in one go (just make all the necessary subfolders) but maybe it's not supported on all browsers, the first one works even from mobile!
>(Note: you don't need to change this PR, I don't mind those extra commits, just saying because probably it's easier for you when uploading files next time)

Thank you for the explanation, I'll look into this for next time, yes I created these test.txt files to empty a folder without it being deleted (it's basic but it works), and also because regularly I "work" directly from my pixel 6 pro (and vanadium isn't necessarily very cooperative)

>You have no idea the amount of work you just saved me.

With pleasure, it took me a while because F-droid is not the rush type (and I waited for V2.6). Thank you for those kind words, I'll follow up when needed (if I don't, contact me).
If you want different (user interface) tints, or other aesthetic changes, let me know.

>There is a way to automate making the screenshots, but I'm still working to make it work

No problem for me to make them, no worries.

>Now I only need to automate uploading them to Play Store, which is also a work in progress (currently I'm doing it manually).

So sorry for the extra work and good luck for that, on my fork of your app, I couldn't even get him to build me an apk for each new version

>Don't worry about the Turkish locale, as you said it will be possible when I add the ability to change the locale in-app.

Awesome, thank you. <3

>Btw, I noticed that the open text in the ukranian locale touches the borders, which means that can happens on all the other locales too. I'll add a bit of padding to fix that.

I only saw this for Ukrainian, it's strange (is it due to the alphabet?)

>Also I need to remove those old [[Fix||http]] tags. Oops

The worst thing is that I only noticed it thanks to your comment (and your recent issue), when this is fixed, I will change the capture, of course ;-)

TrianguloY:
The changes should be ready now.

In the new alpha you can change the locale by clicking the app icon (from the main screen). If it doesn't work or something is wrong just tell me.

Ilithy:
I tested, it works perfectly, and it will make my "work" easier, thanks.

I couldn't make all the screenshots today (I don't have a pc, only my pixel 6p), but it will be done tomorrow, the (Ukrainian) screenshots will be modified this morning

TrianguloY:
Thanks! But don't worry about having everything instantly. Take a much time as you need (I do!)

PR #59, opened 2022-08-21
Issue templates

Ilithy:
Proposal to add (simple) issue templates


TrianguloY:
This is just one of those features I was aware that existed but didn't cared to implement. They seem easy and useful!

Just removed the 'good first issue' from the bug template (not all bugs can be easy to solve!) and merged!

PR #60, opened 2022-08-21
Add funding

Ilithy:
Added funding.yml file _(it doesn't hurt)_
- If possible _(and if you want)_, you should register for the Github Sponsor program _(the funding.yml file is already set up for it)_

### DIFF
--- a/.github/FUNDING.yml
+++ b/.github/FUNDING.yml
@@ -10,6 +10,4 @@ liberapay: # Replace with a single Liberapay username
 issuehunt: # Replace with a single IssueHunt username
 otechie: # Replace with a single Otechie username
 lfx_crowdfunding: # Replace with a single LFX Crowdfunding project-name e.g., cloud-foundry
-custom: ['https://www.buymeacoffee.com/TrianguloY'] # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
-custom: ['https://www.paypal.com/donate/?cmd=_s-xclick&hosted_button_id=W5CZR2YTVT5BQ']
-custom: # Replace with up to 4 custom sponsorship URLs e.g., ['link1', 'link2']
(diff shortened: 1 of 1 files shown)


Ilithy:
Please enable the display of the donate button in the repo settings.
If you don't like it, cancel it with apologies

TrianguloY:
I tried to register for the program in the past, but the amount of personal data they require was high so I never finished it.
Since then I've need to give that data to the other payment providers (Paypal, Patreon, etc) so I guess it's common (and understandably necessary).
I finished the process, just pending for the approval. Let's see if it was worth it!

TrianguloY:
Btw: remember to delete merged branches! (unless you plan to use them later)

Ilithy:
Thank you, yes personal data remains the sinews of war, thank you, I may drop Patreon for Github Sponsors when this famous validation will be done (there is no fee below 5000€ of donations per year if I'm not mistaken, it's more interesting than Patreon, which takes a nice commission so it's less money for you)

>Btw: remember to delete merged branches! (unless you plan to use them later)

Thanks for the reminder, I will, thank you.

Merci pour le rappel, je le ferais, merci.

TrianguloY:
> I may drop Patreon for Github Sponsors when this famous validation will be done

No need. I prefer to give users a choice even if it's worst for me.

Ilithy:
> Ah! You mean yours! As you wish then 👍

I mean to support you (unless you prefer Patreon), github is (it seems to me) more profitable, but there is nothing really problematic

Ilithy:
> Ah! You mean yours! As you wish then 👍

I mean to support you (unless you prefer Patreon), github is (it seems to me) more profitable, but there is nothing really problematic

TrianguloY:
Patreon has a commission, GitHub doesn't, yes. I set the same tiers on both cases, but as I said the ultimate choice is yours :)

Ilithy:
@TrianguloY just a detail, could you set up a recurring donation of 10€ (it's not really more than 7€ I know), thank you.

TrianguloY:
I think I enabled the ability to specify a custom amount. Isn't it possible? 
The tier is automatically adjusted in that case.

Ilithy:
Sorry, it's possible (I just missed it), this validates well at the amount selected

PR #62, opened 2022-08-22
Uk-UA screenshots

Ilithy:
Changes two Ukrainian screenshots (of the dialog)


TrianguloY:
I see you discovered the squash and merge option. It is a wonderful option to keep the commit log clean, specially when using the web to upload/delete individual files (it basically creates a new commit with all the changes altogether).

I wasn't going to say anything because it's not a big problem having individual commits, but I do in fact prefer this squash approach to avoid cluttering the log when possible, it even skips the merge! (the readme pr for example had quite several changes...I should've used it there, oops).

Anyway, good job!

Ilithy:
Yes indeed it's a very convenient option, I usually use this, (a dev I know asked me to do this, I've continued since) and I admit to double check the changes it's much easier.
Thanks

PR #64, opened 2022-08-22
(Proposal) PRIVACY POLICY

Ilithy:
Proposal to add a _(simple)_ privacy policy.

_Can (eventually) be added to the About Url Checker page (show as popup), but that's another story._

### DIFF
--- a/PRIVACY POLICY.md
+++ b/PRIVACY POLICY.md
@@ -10,6 +10,9 @@ Url Checker does not contain any advertising sdk, nor tracker of the user, or hi
 
 Cookies are not stored at any point. Authentication credentials can be stored optionally on the user's local device upon the user's explicit request. developer does not have access to any such information.
 
+All external interactions require user action (to press a button), unless explicitly configured to automatically do so, which is always disabled by default.
+In any case, only the displayed url will be fetched/sent.
+
 ## Third party cloud service dependencies
 
 Note that Url checker:
--- a/PRIVACY POLICY.md
+++ b/PRIVACY POLICY.md
@@ -1,17 +1,16 @@
 
 ## Introduction
-This privacy policy covers the use of the 'Url checker' Android application.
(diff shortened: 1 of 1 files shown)


TrianguloY:
Wow, to think that the current privacy policy is this: https://triangularapps.blogspot.com/p/url-checker-privacy-policy.html

Did you used a template or something? It looks very professional (and it may help me in other apps). Three are a few ideas that could be added present in the link above:

- See the suggestion about the automatic feature
- When opening an app the url, and only the url is sent.

Ilithy:
Honestly I started with a privacy policy of an application (Nova video player) and it is adapted to Url checker, I find it simple and leaves no doubt (except this part pointed out by your comment)

thank you for giving me the link to your privacy policy, I had not seen it --'


Indeed, the part on the automation and the sending of the url are important points, I modify, that in my proposal.

You are free to use it as you see fit.

Ilithy:
Isn't this formulation problematic?

"In any case, only the displayed url will be fetched/sent."

Doesn't the scan module send other data? (such as the user's api key), doesn't that conflict?

TrianguloY:
Hmm, the VirusTotal api key is used to authenticate the petition (if it's wrong it just fails). I treated it as a token, but I'm no lawyer so no idea if it's considered a private data, although I guess it is.
I updated the policy, but I'll replace it with this one so don't worry much about it.

Ilithy:
That's fine with me

TrianguloY:
I decided to remove that sentence. The rest of the document is very detailed and I think it's more confusing that not.
But if you think it's better to leave it I have no problem.

I also added links and tweaked a few punctuation

Ilithy:
All your modifications are excellent, with them it seems to me that it covers everything.

TrianguloY:
Merged then! 
Of course it can be updated further if needed.

PR #65, opened 2022-08-23
Update CONTRIBUTING.md

Ilithy:
(first step) added primary help for translators

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,11 +13,11 @@ You can check existing code to see examples, but be aware that some are not perf
 
 you are free to propose one or more translation(s). Be sure of your work before providing it to the developer, who cannot verify it.
 
-- It is necessary to translate the `strings.xml` file present in the `app/src/main/res/values/` directory
+- It is necessary to translate the `strings.xml` file present in the `app/src/main/res/values/` subdirectory
 
 >- Translate all texts between the `<string>` and `</string>` tags.
 >- Remember to insert a backslash `\` before any apostrophe `'` Or enclose all text in quotes `"`
 
 Optionally you can:
 - Translate the `full-description.txt` and `short-description.txt` files _(presentation for application stores)_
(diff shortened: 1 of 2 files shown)


Ilithy:
I don't know if the way I added this will suit you, I'm thinking of doing something more complete (if it remains, I'll probably make it more exhaustive)

TrianguloY:
I would like to setup an automatic action to generate a zip with the translation files directly (like I have in the Click to chat app). There are also public services for translations that I can check (to avoid manual translation of string files).

If you have time and want to, update as you please, if not don't worry.

Ilithy:
>There are also public services for translations that I can check (to avoid manual translation of string files).

I think this technique would be the most viable, and the most practical (once implemented), have you looked at [Weblate](https://weblate.org/)?
seems like a great compromise, and they have a [free plan for FOSS software](https://hosted.weblate.org/accounts/login/?next=/hosting/).

I'll leave what I added (if you don't mind the edits) and once the final solution is in place I'll remove it, when do you think?

TrianguloY:
> have you looked at [Weblate](https://weblate.org/)?

That's exactly the service I have on my TO-DO list to check someday...

TrianguloY:
I added links and a bit more description to translations.
And also made sentences in plural (I'm the main and currently only developer, but who knows in the future). I also like to have general instructions related to a 'team'

Check them if necessary and do an Squash and merge yourself (I want to check what happens when there are multiple authors)

Ilithy:
Please check in return

TrianguloY:
Note: I change some repository settings:
- When doing a Squash-merge with a PR, the commit that is generated (with all those changes) will have the title and description of the PR itself (currently it was auto-generated). You now have easy control of that commit.
- I enabled the 'always suggest updating the pr branch' option. This should mean that now you can easily rebase a branch to 'main' via web (if it doesn't work as expected and it's more confusing I can remove it again)
- Merged branches are now automatically deleted. A button to restore them will be present if needed.

PR #67, opened 2022-08-23
Set application name as untranslatable

Ilithy:
Set application name _(Url checker)_ as untranslatable

- avoid errors and problems with the translation of the name.
- harmonize the display of the name.
- simplification


TrianguloY:
Hmm. I want the app name to be translatable.

And I was writing a long and boring explanation of why. But the more I wrote the less sense it made, so maybe you are right.

The short explanation is that the current name is not a brand name but a description (Url checker, literally a checker for urls...yeah I'm _that_ original) and as a description it should be translatable, as opposed to original names like Google, Amazon, etc...even though they are existing English words which are so popular that the brand name is independent now.

I was trying to find examples of other description names, but except for some strange cases (Google translate and Music are translated into Spanish at least) almost nothing else is translated. Even on games the original English title is kept, and a translated subtitle is used sometimes if necessary.

This is not usually the case if the original language is not English (specially with non-latin alphabets like Cyrillic or Asian countries). English is truly the universal language...


In conclusion: I don't know what to do. It's probably better to keep the original name as you suggest, even if it means absolutely nothing to a German user for example. Maybe on non-latin alphabets it makes more sense (imagine if you are scrolling F-droid and you see a name in Arabic, that's more or less what arabian users feel when they see an English one, although as I said English is truly the universal language so the comparison is not fair).



I have a lot of doubts on this one, just keep it open for now. I would like to see your (or anyone else) opinion on this topic though.

Ilithy:
Sorry for the delay, I thought long and hard about the question before answering.

one thing became clear to me, it's your application, and if you prefer the name to be translated into the different languages ​​so as to be completely understandable (a descriptive name, as you call it), that's how what it will be.

My opinion is rather general, it is much easier to find an application if its name does not change from one language to another;
**Ps**: except for specific cases, like Google translation, but it's perverse, it's normal that the word "translate" is translated, you see the idea.

This also prevents a translator from deciding (at once) the name of the Url checker (therefore the name of YOUR app), but also avoids misunderstandings...

But feel free to reject this, it's not a decision of value, but of taste, and it can't be judged.

TrianguloY:
Thanks for your opinion.
I've been thinking and I had an idea. If the problem is that the name is descriptive, why not changing it to a non descriptive one? Not necessarily a random sequence of letters, but something like ClearURL.

In fact (and because changing brands is a risky move because you can mislead users) I've thought of two possible new names:
- UrlCheck
- CheckURL

The first one is basically the current name, the second one is ClearURL but with Check.

Any other possible suggestion? Thoughts? Opinions?

Ilithy:
URLCheck seems fine to me, I allow myself to do in the proposal, what do you think of:
Link Detective?
Vinculum Validator? _(hmm a bit too much)_
the Link Eviscerator? _(I'm joking)_
_You will probably prefer a fairly short name..._

Ilithy:
Another idea that came to me (I really like it) is **Linkus** (Link-us), What do you think?
another application has this name _(but it does not own it, not used for url checking and is only available for IOS in OS mobile)_

TrianguloY:
Hehe, Link Eviscerator, that gave a bit of a chuckle :P

Hmm, I personally prefer URLCheck. 'Linkus' is nice, but it's very different so it will probably cause much more confusion.

I think URLCheck is the perfect solution. But let me create a poll first because changing the files.

Ilithy:
Confusion? Why? (Just to know).
Many applications have a name radically different from their utility (the application "[Les Pas](https://github.com/scubajeff/lespas)" (name in french) for example, which is an album for Nexcloud) Link eyes is another example (wink)...
For me, the goal is to have a name that pleases you and the user. If your preference is UrlCheck, it's up to you, for sure.

The intro text and the keywords are enough most of the time to find an application on the app stores

Let's hope that your survey will be seen by your users

Ps: I will not merge anything without your explicit validation 👍

TrianguloY:
Confusion for existing users, mainly. 

In any case I can't say exactly why, but I like the 'check' in the title, one of those inexplicable personal preferences I guess 🤷

Ilithy:
I like its current name, is the initial reason for this PR, keep it identical in all languages, so Url Checker or UrlCheck suits me, I would not add anything your honor ^_°

TrianguloY:
OK, a week has passed and the poll says that the winner is "URLCheck" (there are only 4 votes though, but anyway).

You now have my permission to update this PR with that new name. It would be a good idea to update the Readme and other places where the name appear, but this will happen for the next version so there is no rush.

TrianguloY:
Is this ready to merge or are still some files you need to update? 
(You assigned me to the pr, great! But you assigned yourself too so I'm not sure if you still want to do something else 🤔)

Ilithy:
Yes it was to imply that you can look at my work (it seems ready to me), but that if needed, I will make the changes, it was a bit obscure as an idea, sorry

Ilithy:
The header image will have to be changed quickly in the near future, it's up to you to tell me if you want me to do it or you take it in hand

TrianguloY:
You already set me as reviewer, which at least for me imply that you want me to review the changes.

In the end we can use whatever options we want (and I'm open for suggestions) but for me:
- the reviewer field is for people that should check the pr (as you are asked doing), can be multiple
- the asignee is for the person that will merge it (and are in charge of it), usually just 1.

In this case I would set you as the asignee (it's your pr and you are going to merge it) and me as reviewer to check the changes. Or you can set me as asignee too, if you prefer me to merge it.

TrianguloY:
> The header image will have to be changed quickly in the near future, it's up to you to tell me if you want me to do it or you take it in hand.


Yeah, I can do that, don't worry

[Oops, I mistakenly edited your post instead of replying to it. Sorry! 😅]

Issue #39, opened 2022-08-04
Configure custom tabs initial state


TrianguloY:
Currently the custom tabs button is always enabled and the initial state correspond to what the original app configured. Ideally you should be able to change this behaviour. I'm thinking (as a possible solution, but may be others) of adding a dropdown in the open module settings with the following options:

Custom tabs:
- 'Auto' -> The current behavior, default selection.
- 'Initially on' -> The custom tab button is 'on' when shown (you can manually turn it off)
- 'Initially off' -> The custom tab button is 'off' when shown (you can manually turn it on)
- 'Disabled' -> The custom tab button is not shown. The custom tab feature is always off.

Issue #37, opened 2022-08-04
Some weird interactions in the Open & Share module


PabloOQ:
When I tap on a link in the twitter app and URLChecker opens, with custom tabs 'off', if I then open the link in Firefox Focus or the DuckDuckGo browser with these apps closed, the browser opens 'inside' the twitter app, this doesn't happens with other browsers. Because of this behaviour, if I then go to another app the browser inside the twitter app closes, and I need to reopen the link. I haven't seen this happening in other apps either, only on twitter, which makes me think this is something that can be solved inside URLChecker. It would be great to have an option to avoid this, maybe a button, or a preference in the Open & Share module. Maybe even have something added to the custom tabs icon to show this, and change the behaviour by holding the button.
Some screenshots of the button concept:
- No custom tab, open externally
![image](https://user-images.githubusercontent.com/55493443/182905231-e0c055aa-1eea-4e70-90ce-300adc36420c.png)
- No custom tab, open internally
![image](https://user-images.githubusercontent.com/55493443/182905286-fd891ebd-38cc-4320-adcc-c5aeb5283284.png)
- Custom tab, open externally. 
![image](https://user-images.githubusercontent.com/55493443/182905161-f849caed-09ab-4016-aa0d-705abe4d056f.png)
- Custom tab, open internally.
![image](https://user-images.githubusercontent.com/55493443/182905132-04378912-3556-4bdc-8dc6-9bd3e8fc985b.png)

Some screenshots of the behaviour:
- URL Checker
![image](https://user-images.githubusercontent.com/55493443/182902318-830d4ef5-8039-42b2-8aaa-7557677b6388.png)
- While going to change to another tab (before the animation concludes), the "inside browser" is still open
![image](https://user-images.githubusercontent.com/55493443/182902382-7d5d33f3-c70b-413f-a2a5-c9e3302d58f5.png)
- When going to change to another tab (animation concluded), the "inside browser" is closed
![image](https://user-images.githubusercontent.com/55493443/182902449-93c182c4-9ec5-46f6-8859-704e0eb10620.png)

Also when I tap a link in the youtube app the custom tabs button is 'on' by default, but on other apps is usually 'off', the twitter app for example.  Same thing as the previously mentioned, a preference to choose the default state, 'on' or 'off' would be great, currently if feels incosistent accross apps.

TrianguloY:
2 issues here:
[...]

Issue #47, opened 2022-08-11
Option to show all apps that can open a link (UI improvement)


advait404:
Firstly, thanks a lot for this wonderful app. its by far the best URL cleaner on android. 

I use the [Open Link With](https://f-droid.org/packages/com.tasomaniac.openwith.floss/) app that instantly shows all apps that can open a link. So it takes only 2 clicks: one to select the open link with app, and one to choose the specific app to use.

however, In URL checker there is only 1 app shown that can open the link, and the rest are accessed by clicking on the arrow. This takes 3+ clicks and is a bit confusing. I think an option to change the view from showing 1 main app into a list format will be nice. It will make opening pprcess much quicker, and with the automatic URL cleaning enabled, its   a very simple and quick process. 

Additionally, it would be great if the apps' logos are also visible next to their names in the list, making identifying the apps easier and faster.

TrianguloY:
I think I mentioned it somewhere, but funny fact my original prototype had exactly this, a list with icons (a _very ugly_ list).
I'm not really sure why I changed it, maybe because with the module approach it took a lot of space or perhaps the share button was hard to add in that case.

I've been thinking into adding a setting to toggle between the button+arrow and icons list too, and it would in fact be a good addition.

So many good ideas, so little time...

advait404:
Wow, thanks for the super fast reply. I wish I could help you more with this, but I barely have any coding experience.

Frankly, the current UI is really good and works well. So I don't think this is a priority, but it would be great to add.

One 'easy' way of implementing this could be to show the app list by default, and to show this list below the arrow instead of above it. This way, the user can click on the app directly. It probably won't look great, but works decently well.

Is it possible to maybe copy code from the open link with app and reuse it here? Or maybe some collaboration between the projects? It could help with implementing the best of both apps.

TrianguloY:
Moving to an inline list, is easy. Making sure that it looks good and it's usable isn't. (and also testing that it doesn't break anything else!)
I'm planing to separate the open and share modules, that could also help to give more space...

Efreak:
Try looking at the way [link eye](https://gitlab.com/kuesji/link_eye) displays a list of apps.

I'm not a fan of the current color styling of link eye, so I use an older version, however I believe the list and most of the button arrangement is the same. I really link the way link eye provides the 4 buttons at the bottom, before the list of apps: copy, share, clean, close. 

[...]

PR #68, opened 2022-08-23
Open and share update

### LINKED ISSUE #14: Can´t check URLs from YouTube
Prerequisites:
- Default browser app set to none

Steps to reproduce:
1. Open YT app.
2. Go to any video with links in the description. [E. G.](https://www.youtube.com/watch?v=wSy3kAHtmF8)
3. Click any non-YT link.
4. UrlChecker doesn't show up in the "Complete action using" pop-up dialog.

Note that the URL it gives you is a YT redirection, so any links to social media pages won't give you the option to open it in their app, which can be easily solvable with the ClearURLs module.

The only apps I can get to show up are:
- Brave browser.
- DuckDuckGo browser.
- Kiwi browser.
- Firefox browser.
- Firefox focus browser.
- Tor browser.
- URLSanitizer, which is an app to handle URLs too.

Other apps like "Share to clipboard", "UntrackMe" or "Open Link with" do not show up either.

From what I can see it looks like UrlChecker needs to show up as some kind of browser for this to work.

PabloOQ:
Fix #14 
First commit of the branch, easy fix of the manifest.

[Added Integer preference](https://github.com/TrianguloY/UrlChecker/commit/a49d6d04927196737a79f9d080d9088fc5490ecd)
I didn't reuse long because the item selected on the spinners give an int not long. This should be of use to attach any spinner to a preference, could be useful to choose the language as a preference, or select night mode.

Implemented #39 

Suggested solution for #37
Preset flags feature: There are a number of presets flags to apply to the intent, overriding all previous flags. To make space for the button a dropdown was added, which contains any additional settings of the module that can be modified in the dialog. There are 2 presets at the moment:
- Temporal: Works as the [twitter video 'inside' ](https://github.com/TrianguloY/UrlChecker/issues/37#issuecomment-1205743582). Doesn't always work, depends on the app, it didn't work on Brave for me, but it did on Firefox focus, YT and Twitter for example. 
- New window: Works as the [youtube video 'outside' ](https://github.com/TrianguloY/UrlChecker/issues/37#issuecomment-1205743582)
There are 3 icons added, the `mopen_auto.xml` icon was just copy paste from other icon as I didn't know what it should be. When the feature detects that the intent is the same as one of the presets it will not show the auto icon while rotating the presets in the dialog.

I intentionally added the preset flags feature the last so it can be easily reverted or entirely ditched, as I must admit the solution is somewhat weird. Feel free to remove it if you feel it doesn't fit the app.

Each commit has it's spanish translation too if there was any string added.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -39,6 +39,17 @@
                 <data android:scheme="https" />
             </intent-filter>
             <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.BROWSABLE" />
+                <category android:name="android.intent.category.DEFAULT" />
+
+                <data android:scheme="http" />
+                <data android:scheme="https" />
+
+                <data android:mimeType="text/html" />
+            </intent-filter>
+            <intent-filter>
                 <action android:name="android.intent.action.SEND" />
                 <category android:name="android.intent.category.DEFAULT" />
                 <data android:mimeType="text/plain" />
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/OpenModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/OpenModule.java
@@ -23,16 +23,44 @@ import com.trianguloy.urlchecker.modules.AModuleData;
 import com.trianguloy.urlchecker.modules.AModuleDialog;
 import com.trianguloy.urlchecker.modules.DescriptionConfig;
 import com.trianguloy.urlchecker.url.UrlData;
+import com.trianguloy.urlchecker.utilities.GenericPref;
 import com.trianguloy.urlchecker.utilities.LastOpened;
 import com.trianguloy.urlchecker.utilities.PackageUtilities;
 import com.trianguloy.urlchecker.utilities.UrlUtilities;
 
+import java.util.LinkedList;
 import java.util.List;
 
 /**
  * This module contains an open and share buttons
  */
 public class OpenModule extends AModuleData {
+
+    /**
+     * Values for the config
+     */
+    enum CtabsValues {
+        AUTO(Key.AUTO),
+        ON(Key.ON),
+        OFF(Key.OFF);
+
+        public final int key;
+
+        CtabsValues(int key) {
+            this.key = key;
+        }
+
+        // Values to be stored
+        private static class Key {
+            public static final int AUTO = 2;
+            public static final int ON = 1;
+            public static final int OFF = 0;
+        }
+    }
+
+    public static GenericPref.Int CTABS_PREF() {
+        return new GenericPref.Int("open_ctabs", CtabsValues.AUTO.key);
+    }
 
     @Override
     public String getId() {
@@ -56,7 +84,7 @@ public class OpenModule extends AModuleData {
 
     @Override
     public AModuleConfig getConfig(ConfigActivity cntx) {
-        return new DescriptionConfig(R.string.mOpen_desc);
+        return new OpenConfig(cntx);
     }
 }
 
@@ -65,6 +93,8 @@ class OpenDialog extends AModuleDialog implements View.OnClickListener, PopupMen
     private static final String CTABS_EXTRA = "android.support.customtabs.extra.SESSION";
 
     private LastOpened lastOpened;
+
+    private final GenericPref.Int ctabsPref = OpenModule.CTABS_PREF();
     private boolean ctabs = false;
 
     private List<String> packages;
@@ -76,6 +106,7 @@ class OpenDialog extends AModuleDialog implements View.OnClickListener, PopupMen
 
     public OpenDialog(MainDialog dialog) {
         super(dialog);
+        ctabsPref.init(dialog);
     }
 
     @Override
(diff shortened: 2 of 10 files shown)


TrianguloY:
A quick glance over the code and everything seems ok, although I'll refactor a couple things as usual.
Nice commits, nice comments, bravo!
Just give me a couple days to correctly review it.

I would have made the #14 fix a separate PR though (it's a different thing). Similar with #39 and #37 but those are related so it's ok to keep them altogether. 
In any case don't worry, no need to separate them now (although I think you can create a branch from that first commit, make a PR for it, and then this one will be automatically updated without it...hmm, you can try if you want).

PabloOQ:
I will keep that in mind for future PRs

> I think you can create a branch from that first commit, make a PR for it, and then this one will be automatically updated without it

[Like this?](https://github.com/PabloOQ/UrlChecker/commits/Fix-manifest)

TrianguloY:
> I will keep that in mind for future PRs

Perfect, also thanks for taking the time to group the commits nicely 👍


> [Like this?](https://github.com/PabloOQ/UrlChecker/commits/Fix-manifest)

I think so. Try creating a PR from that branch!

TrianguloY:
Hmm, it didn't work :/
I guess you should rebase the branch for it to detect the changes? but that can make a lot more problems so better not. It's just a single file, no big issue.

PabloOQ:
I am using the wrong flags on `EXCLUDE_RECENTS`, the purpose of _New window_ is to open it externally which currently does not always do. As you mentioned on #37, I think the flag which should be used is `FLAG_ACTIVITY_NEW_TASK`, I will do more tests on this. Which leads to something I suspected I was doing wrong, using the flags as keys.
Also the icon for `AUTO` is not the correct one. 

> When the feature detects that the intent is the same as one of the presets it will not show the auto icon while rotating the presets in the dialog.

Lastly, I think it would be better if the button has a different color whenever the current option corresponds to the original flags of the intent.

I can fix these issues, but how should I proceed? I can close this PR and split it in 2, the custom tabs buttons and the preset flags, or I can just do the commits in this PR.

TrianguloY:
Wait. I've started working on this PR. I'll let you know later when I have time to write more than a few sentences 😅

TrianguloY:
Ok, lets go.

I started reviewing the changes. The first thing I saw is that, as you said, the preset flags feature is a bit strange. Ideally there should be only 1 flag, or multiple but independent. Also '[FLAG_ACTIVITY_BROUGHT_TO_FRONT](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_BROUGHT_TO_FRONT)' is marked as 'not normally set by application code'.
The icons must also need to be changed (maybe even setting text next to them to clarify), so I decided to skip that commit and only merge the #39 change.
In any case your solution for #47 could be useful in the future (I really like the dropdown with the extra options there), I can save it on a branch in this repository if you want.

As for the #39 feature, as usual I started changing and refactoring code (this is my fault, sorry). Mainly the way the preference is saved because you can either save the enum index or the enum name. With an index you depend on the enum order itself, which can change under some very specific situations. With the name you can't rename the enum values, but at least you can reorder/add/remove new elements if needed. I don't know if there is a better way, but I prefer saving the enum name instead of your index choosing. 
In any case I'm refactoring it so that the 'preference' is the enum itself! So that the code using it works with the enum directly (no matter how it's saved) and the preference class is the one deciding how to save it (it's a better separation of responsibilities, I think). I'm also adding a way to attach a translatable string to any enum value directly, instead of using a separate pair-like class.

Hopefully this afternoon I'll be able to upload it. You don't need to do anything for this, but if you want try to check if the FLAG_ACTIVITY_NEW_TASK is enough for the #39 feature.

<br>

Btw...as I asked Ilithy before: Would you like to become a collaborator? [In your case as a programmer I'll probably ask you to create a PR for any non-obvious change anyway, but your help fixing small bugs or even translating spanish could be useful.]

PabloOQ:
> Also '[FLAG_ACTIVITY_BROUGHT_TO_FRONT](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_BROUGHT_TO_FRONT)' is marked as 'not normally set by application code'.

I just took the flags that twitter uses to mimic the 'temporal' behaviour.

>  Mainly the way the preference is saved because you can either save the enum index or the enum name. With an index you depend on the enum order itself, which can change under some very specific situations. With the name you can't rename the enum values, but at least you can reorder/add/remove new elements if needed. I don't know if there is a better way, but I prefer saving the enum name instead of your index choosing.

That's exactly why I used a key, using the index was a big no, and using the string while better wasn't really a good alternative.

> In any case I'm refactoring it so that the 'preference' is the enum itself! So that the code using it works with the enum directly (no matter how it's saved) and the preference class is the one deciding how to save it (it's a better separation of responsibilities, I think).
That's what I tried at first, but couldn't get it to work, that's when I decided to rely on the key value class. Hopefully you get it to work.

> Btw...as I asked Ilithy before: Would you like to become a collaborator? [In your case as a programmer I'll probably ask you to create a PR for any non-obvious change anyway, but your help fixing small bugs or even translating spanish could be useful.]
I don't mind, in the end is what you are saying, I would still do PRs for features, but I can do translations when needed.

TrianguloY:
It's merged now! (the #39, not the #47). I even added two extra settings: always on (the button is hidden but the ctabs are enabled) and always off (same but with off).

> That's exactly why I used a key, using the index was a big no, and using the string while better wasn't really a good alternative.

I investigated and saw that using an extra integer, exactly what you did, is the correct alternative in almost all cases (turns out that under some conditions the enum values can change, mainly with proguard and obfuscation). With my refactor I think it even looks nicer.

> That's what I tried at first, but couldn't get it to work, that's when I decided to rely on the key value class. Hopefully you get it to work.

I hope so! You can check it and see if you like it.

> I don't mind, in the end is what you are saying, I would still do PRs for features, but I can do translations when needed.

Ok! Just let me prepare a few things first and I'll send you the invitation.

TrianguloY:
I'm closing this PR as it is no longer needed. I keep the extra commit in shelf/PR68 if needed.

PabloOQ:
It's exactly how I wanted to do it! It's a shame I couldn't make it work, I apologize for the inconvenience.

TrianguloY:
I admit it was hard to manage to make it work properly, but I liked how it ended...except for the TranslatableEnum (why on Java you can't extend enums...why...on Kotlin you can...if only Kotlin weren't so badly optimized...).

Anyway, as you probably saw I use this and other free-time projects to allow making interesting and 'innovative' (at least for me) programming things. I had a great time making this, no need to apologize.

I'll even probably use this opportunity to learn about annotationProcessors. Let's see if I can make one!

PR #70, opened 2022-08-23
Fix #14

### LINKED ISSUE #14: Can´t check URLs from YouTube

PabloOQ:


TrianguloY:
I probably though that by adding no mimetype it would match all, but apparently isn't?

I wonder if adding "*" as mimetype will also work and allow even more intents to work.

In any case it's just a very minor fix that adds an intent, and if you tested it and it solves the bug then perfect!

PabloOQ:
Honestly, the bug itself was very specific, as after setting URLChecker as the browser it worked correctly, but if it wasn't then it wouldn't show up only in the youtube app. I wouldn't look too much into it.

Ilithy:
> I probably though that by adding no mimetype it would match all, but apparently isn't?
> 
> I wonder if adding "*" as mimetype will also work and allow even more intents to work.

I am not totally sure of myself, but there may be a solution to this, I attach an [issue](https://github.com/Kunzisoft/KeePassDX/issues/1211#issuecomment-1218456465) in which the developer @J-jamet seems to have a similar problem, it may help..._(maybe)_

TrianguloY:
I think [this](https://developer.android.com/guide/components/intents-filters#DataTest) is the important part:

By no specifying mimetype some cases don't match, apparently YouTube is one of them.

If another similar issue arises I'll check again, but for now this is a perfectly good solution. 👍

PR #71, opened 2022-08-24
Iw screenshots

Ilithy:
* Added screenshots in Hebrew (first step)

* The Hebrew language code is no longer iw-IL but he-IL, see ->https://www.localeplanet.com/icu/he/index.html
_I didn't touch it, should we change this?_


TrianguloY:
On Play Store hebrew is iw-IL, I guess f-store keeps the same. No need to change anything.

Ilithy:
Thanks 🙏

PR #77, opened 2022-08-27
Update gradle to 7.5.1

Ilithy:


TrianguloY:
Don't worry about gradle updates, Android Studio (the program used to develop the apps) automatically notifies me when a new version is available, and updates the necessary files itself. The Android gradle plugin also must be sync with the standar gradle, so not all versions can be used (I think).
 https://developer.android.com/studio/releases/gradle-plugin 

I'm going to reject this update because I prefer to let Android Studio update and manage the synchronization.
Thanks in any case!

Ilithy:
No problem, indeed, it's probably safer this way, thanks for the explanation.

PR #78, opened 2022-08-27
Build test action

Ilithy:
Hi;
**You are totally right** Github actions is very funny, I hope not to disturb (I just suggest, for fun, tell me if it annoys you, I'll keep it on my fork, warm ^_*)

* This workflow that performs continuous integration
 (CI) for your Java project.
* Will allow you to see when commits to a pull request cause build or test failures against your default branch; this approach can help ensure that your code is always healthy. You can extend your CI workflow to cache files and upload artifacts from a workflow run.
* ~~See: https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-maven~~

### DIFF
--- a/.github/workflows/validate-store-files.yml
+++ b/.github/workflows/validate-store-files.yml
@@ -1,19 +1,21 @@
-name: Validation
+# This action checks that the store files are valid for Play Store. Current checks:
+# - title and description files must not exceed a maximum number of characters
+# To be replaced with a more powerful (and less manual) one
+name: Validate store files
 
-# Note: this is the second time I'm writting a GitHub action, and the first time I'm using a script.
+# Note: this is the second time I'm writing a GitHub action, and the first time I'm using a script.
 # Most probably there are better ways to do it, if you know one and want to suggest it that'd be wonderful!
 
 on:
   push:
     branches:
-    - master
+      - master
   pull_request:
-    branches: 
-    - master
-    
+    branches:
+      - master
+
 jobs:
-  validate:
-    # This action checks that the title and description files don't exceed the maximum number of characters allowed on PlayStore
+  validation:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
@@ -23,20 +25,20 @@ jobs:
             // import
             const { lstatSync, readdirSync, readFileSync, existsSync } = require('fs');
             const { join } = require('path');
-            
+
             // header
             console.log("Validating files:");
             console.log("");
-            
+
             let error = false;
-            
+
             // for each locale
             const listings = "./app/src/main/play/listings";
             readdirSync(listings)
               .map(locale => [locale, join(listings, locale)])
               .filter(([locale, path]) => lstatSync(path).isDirectory())
               .forEach(([locale, path]) => {
-                console.log(`[     ] - '${locale}'`);
+                console.log(`[      ] - '${locale}'`);
 
                 // check files length
                 ([
@@ -44,7 +46,7 @@ jobs:
                   ["short-description.txt", 80],
                   ["full-description.txt", 4000],
                 ]).forEach(([file, limit]) => {
-                
+
                   // get file
                   const subpath = join(path, file);
                   if(existsSync(subpath)){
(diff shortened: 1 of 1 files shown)


TrianguloY:
Exactly ;)
This is more or less what I did for the alpha action, told you it was fun!

Just one thing: the link you mention is incorrect, Android uses gradle, not maven, I guess you missclick when copying it and wanted to link [building-and-testing-java-with-gradle](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-gradle) instead?
On that link it is also mentioned that you can use the [gradle-build-action](https://github.com/gradle/gradle-build-action) action for some benefits over the short-but-manual approach.
Maybe you can set that up? (Should be just changing a few lines)

Ilithy:
> Exactly ;) This is more or less what I did for the alpha action, told you it was fun!
> 
> Just one thing: the link you mention is incorrect, Android uses gradle, not maven, I guess you missclick when copying it and wanted to link [building-and-testing-java-with-gradle](https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-gradle) instead?

Yes I'm sorry I'm doing this from my phone, an unfortunate copy paste

> On that link it is also mentioned that you can use the [gradle-build-action](https://github.com/gradle/gradle-build-action) action for some benefits over the short-but-manual approach. Maybe you can set that up? (Should be just changing a few lines)

I made some changes but I don't know if this is what you want, let me know what you think

TrianguloY:
Well, I see you copied the example on that page. Let's see what it does.

First, it runs the action on all 3 operating systems. This could be a nice option if you are building a multiOS app or want to make sure that something works on all 3 OSes. In this case we are compiling for an external platform (android) and gradle usually takes care of ensuring it can be built independently of the OS, so it's a bit useless to test it on all of them. Only linux is enough.

Second, you added the extra step, but didn't remove the other which is now not needed.

In any case, I think this is a better sample
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-java-with-gradle#using-the-gradle-starter-workflow (the first code under the linked part). It is explained below too, and as you will see it even includes the other wrapper-validation action! So we don't need that one.
The only thing that I don't like is that it's using specific commits instead of versions, we better change those.

To recap: replace with the linked action with the following changes:
- Keep the current "on" part (so it runs on pull requests too)
- Replace the hash with versions. If you are not sure of this step just leave them and I'll change it myself.
- Remove the other wrapper validation action (it will be validated here)


Note: I can easily make it myself, but since you are interested I'll leave you discover it yourself. In any case, if you prefer me to do it just say so.

TrianguloY:
Need some help? 

When I said "the linked action" I was referring to this snippet:

```
# This workflow uses actions that are not certified by GitHub.
# They are provided by a third-party and are governed by
# separate terms of service, privacy policy, and support
# documentation.

# GitHub recommends pinning actions to a commit SHA.
# To get a newer version, you will need to update the SHA.
# You can also reference a tag or branch, but the action may change without warning.

name: Java CI

on: [push]

jobs:
  build:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3
      - name: Set up JDK 11
        uses: actions/setup-java@v3
        with:
          java-version: '11'
          distribution: 'adopt'
      - name: Validate Gradle wrapper
        uses: gradle/wrapper-validation-action@e6e38bacfdf1a337459f332974bb2327a31aaf4b
      - name: Build with Gradle
        uses: gradle/gradle-build-action@67421db6bd0bf253fb4bd25b31ebb98943c375e1
        with:
          arguments: build
```

Ilithy:
> Need some help?

Please, yes :/

> When I said "the linked action" I was referring to this snippet:

Yes, I understood for the link (you guided me well to this), but I'm a little lost, it looks like what I had proposed at the beginning, you probably didn't talk about the intro text of the file. suddenly, I don't know what to do to do well (but I'm trying, I promise), if you don't mind, do what you want, so I'll see what you want (with my apologies for not having successful)

TrianguloY:
> > Replace the hash with versions. If you are not sure of this step just leave them and I'll change it myself.
> 
> You lost me on that one, sorry

I meant to use `gradle/gradle-build-action@v2` instead of `gradle/gradle-build-action@67421db6bd0bf253fb4bd25b31ebb98943c375e1`. On the first one you are specifying to use [the v2 tag](https://github.com/gradle/gradle-build-action/releases/tag/v2), on the other [a specific commit](https://github.com/gradle/gradle-build-action/commit/67421db6bd0bf253fb4bd25b31ebb98943c375e1)
The text recommends using a commit hash, probably because that way you know exactly which code you are running (and no one can change that, the hash is calculated with the files, different files means different hash). With the tag option that repository owner can change it whenever they want, which can be a good thing (automatically use the latest bug-free and problem-free version) or a bad thing (introduce unwanted code or something). Since that repository is the official from gradle with 350+ stars, I think it's better to leave the autoupdate on.

TrianguloY:
Ah, and don't feel bad for not knowing what to do, in fact you did quite well until I started to 'suggest' things 😅. My apologies.

I'm also new with these so called 'actions' but I'm not new to coding so I have quite some more experience to guide me.
Even in that case, I need to test a lot! 
image
(I'm doing the test on a private repository, to avoid polluting this one)

Ilithy:
Thank you very much, for the support, the help and the corrections.
When I see the modifications that you have made, it confirms me in the idea that I was far from what you imagined.

Tell me, I take the opportunity to ask you, are the Code Scan Actions useful? (like Codacy Security Scan or CodeQL) I tried to implement them (with success) and the first one gives me warnings:
https://github.com/Ilithy/UrlChecker/actions/runs/2943269724
_(Just for info)_

TrianguloY:
If they can detect problems with the code I'm all-in for them!

As for those warning, I'm more concerned about the errors in the 'Run Codacy Analysis CLI' step. Maybe it needs some more configuration? I'm not sure what this code scan does excatly, I can try to investigate too.

Ilithy:
That's what I thought too, but I understand nothing on it, fairness (I'm kicking in, it's not my level)

TrianguloY:
Ah! I now see the warnings. I was confused with the ones in the action itself.

Those are java best practices, which I do like to follow (but I haven't enabled them on my computer yet). I'll try to fix them if I have nothing else to do, in any case they are not security issues so it's ok.

Ilithy:
>Ah! I now see the warnings. I was confused with the ones in the action itself.

Ok, I see.

> Those are java best practices, which I do like to follow (but I haven't enabled them on my computer yet). I'll try to fix them if I have nothing else to do, in any case they are not security issues so it's ok.

I can imagine that if the application works well (which is the case), and that there is no major security flaw, there is no real reason to rush on this point indeed, it can largely wait, thank you for looking, and for keeping me informed about this anyway.

PR #84, opened 2022-08-31
Pr auto label

Ilithy:


TrianguloY:
This seems useful, are you still working on it?
Also, the action you are using is an extension on the official https://github.com/actions/labeler. 
I think the official one is even more powerful (and it's usually better to use official actions) in case you want to try to use that instead.

Ilithy:
> This seems useful, are you still working on it? Also, the action you are using is an extension on the official https://github.com/actions/labeler. I think the official one is even more powerful (and it's usually better to use official actions) in case you want to try to use that instead.

Yes, it seems to be very convenient for me too, 
I tried (both, the fork and the original), I don't get any error, but the action doesn't label the PRs, I continue my tests (currently on this branch: https://github.com/Ilithy/UrlChecker/commit/9fad2ba848ed885dd0469e2a1d0b98d734a37000)

PR #95, opened 2022-09-13
Updated pt-PT translation (added play metadata)

ReduxFlakes:


Ilithy:
@TrianguloY I will wait for this update of the Portuguese files to be taken into account to modify the Portuguese screenshots, @ReduxFlakes thank you for this translation work

ReduxFlakes:
Thanks, always helping on open source!

Issue #42, opened 2022-08-05
Option for app to close upon passing of URL via open or share buttons


ig9000:
For at least some use cases, there's no further need for a URLChecker instance after the URL has been processed and passed along to another app using the browser or share buttons. This could be a simple boolean setting, or it could even be conditional based on a regexp or something. It might make sense for "open" and "share" to be individually configured in this regard, especially if they are split into separate modules.

ig9000:
Implementation-wise, maybe it would make sense to make this a checkbox shown in the modules for open/share, so that it could be easily disabled for the (odd?) cases where keeping the app open is desired.

TrianguloY:
The current behavior is to keep the app open, or to be precise not to close it. [There is no finish() call when you press the open button](https://github.com/TrianguloY/UrlChecker/blob/master/app/src/main/java/com/trianguloy/urlchecker/modules/list/OpenModule.java#L242).
However this app is configured to not appear on recents so when you open a different app there is no way to go back and it appears as closed even though it's not (well, android probably closes it, but not the app itself)

ig9000:
Yes, this might make more sense for "share" than for opening in a browser. The share functionality is sometimes used to hand the URL to apps which function in the background (downloader, bookmarking, etc). Those are the cases where I'm then left with URLChecker interface still in front of me, but no longer needed. It's not a big deal, but it would streamline things a little.

Example use case is being sent a URL via SMS while using a laptop or desktop device, and you want to quickly and easily pass that URL to one of the various methods for making it available on that device, while continuing to interact with the SMS app on the phone.

PR #109, opened 2022-09-26
Close on share (for background apps)

PabloOQ:
Implemented #42 
Next step would be to implement a way to look if the share was succesfull or not, this way URLCheck doesn't close if the user unintentionally shares the URL.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/OpenModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/OpenModule.java
@@ -35,6 +35,10 @@ import java.util.List;
  */
 public class OpenModule extends AModuleData {
 
+    public static GenericPref.Bool CLOSESHARE_PREF() {
+        return new GenericPref.Bool("open_closeshare", true);
+    }
+
     @Override
     public String getId() {
         return "open";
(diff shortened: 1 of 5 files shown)


TrianguloY:
Looks good! 
- Ignore the code scanning comment
- I think in the merge you forgot some commits, the "This module can't be disabled" strings should not reappear.
- I'm not sure about it being enabled by default, but I guess it's better this way so ok.
- Are you confident about adding another one for the open? Just the same (calling finish) should be enough?

Ilithy:
_I took the liberty of adding the new string to the FR file, I did not touch anything else_

PabloOQ:
> I think in the merge you forgot some commits, the "This module can't be disabled" strings should not reappear.

Whoops, I was working on an old commit (a month ago or so), and some strings reappeared, thought I cleaned everything.

> Are you confident about adding another one for the open? Just the same (calling finish) should be enough?

Sure! Should I add another preference or reuse the same one?

TrianguloY:
> Sure! Should I add another preference or reuse the same one?

I think it's better to use another

Ilithy:
_The new option is translated into French_

Ilithy:
_The new option is translated into French_

ig9000:
Just now got around to trying this out, works as I hoped it would. Thank you!

Issue #82, opened 2022-08-31
New module: Warn flagged urls from hosts-like files


TrianguloY:
The hosts file is a relatively standard and not very hard to parse format (ip + space + domain). 
These are used on windows and linux (and mac?) to disable connections to specific urls by specifying 0.0.0.0 as ip.

https://github.com/StevenBlack/hosts is a good example of these type of files.

### Suggestion
A new module that parses this file and warns when the url domain matches one of them.

Notes:
- It would be very beneficial to have automatic updates
- It would be useful to be able to specify multiple files, even user strings or even from the local device
- IMPORTANT: these files can be extremely long (100000+ hosts) sometimes even with repeated domains. A simple linear search is NOT a good approach, as it will probably halt or even crash. Binary search is better but the best solution will be a [trie](https://en.wikipedia.org/wiki/Trie) or other similar tree-like structure for constant access [^linear].

[^linear]: Well, it will be linear in the input url, but constant in the number of hosts from the file.

Issue #61, opened 2022-08-21
Privacy redirector json created for pattern checker module


Raj9039852537:
privacy rediredtor extension support is still pending and I have managed to make it work with pattern checker module. Here is the link of my configuration for the same [privacy redirector in json formal as required by pattern checker](https://gist.github.com/Raj9039852537/d12e73c6314c05bc79e2bdd71a1001bc).

For temporary or permanent solution this can be used for redirecting the sites.
Feel free to use and edit the gist mentioned according to the needs

TrianguloY:
This is exactly what I had in mind while adding the advanced json editor, letting users create their own patterns in a (relatively) easy way. Good job! 

I, too, have a few personal patterns myself. I now need to think of a way to allow users to discover and upload/download them...

Maybe it's time to setup the Github wiki 🤔

Raj9039852537:
> Maybe it's time to setup the Github wiki 🤔
 That will be more useful for the project

Ilithy:
it is not possible to integrate the Json file into Url Checker, and let users choose to activate or name such and such a redirection?
By this I mean providing a maximum of alternatives directly in the Json file (for example, for Youtube, "providing" Invidious, but also NewPipe, and others), and providing the choice to activate them or not?
The file might be updated (by the developer, or by the community via PR) during Url Checker updates. And "will not force" users to relatively technical manipulations (ordinary users).

Am I delirious or is this possible?

TrianguloY:
It is indeed possible, and I though about adding them disabled by default. But the names are in english, and having them translatable is the real pain...
A good solution would probably be allowing different urls of remote files with the rules (like the ClearURLs catalog but with multiple), but of course the work required for that is not trivial. Time to time.
In the meantime, copy/paste always work (I may add a link to the wiki when it is configured though, for easy access).

Ilithy:
I can't imagine the amount of work that this entails, I have every confidence in you on this subject.

You are right, it is easier to make it functional by use your idea (provide the Json, and the instructions for modifying the rules, as for ClearUrl), my intervention was clumsy.

On the other hand, if you talk about the names of the various applications, they do not have to be translated (in the majority of cases, the name of the application remains the same, regardless of the language of the user)

TrianguloY:
Hmm, that's right, I can add just 'Youtube -> Invidious'. I'll probably do it.
In any case, a wiki is also needed.

Raj9039852537:
[...]

Issue #10, opened 2022-02-16
[New module] Skip redirect


stephenhawk8054:
Hi, I don't know if it would be too difficult to implement or not but it would be nice if there's another module to skip redirection links, similar to how `Skip Redirect` extension works:

[https://github.com/sblask/webextension-skip-redirect](https://github.com/sblask/webextension-skip-redirect)
[https://addons.mozilla.org/en-US/firefox/addon/skip-redirect/](https://addons.mozilla.org/en-US/firefox/addon/skip-redirect/)

For example this link

```
https://www.google.com/chrome/?or-maybe-rather-firefox=https%3A%2F%2Fwww.mozilla.org/
```

can be rewritten to

```
https://www.mozilla.org/
```

I really like this project, one of my must-install app on android now.

TrianguloY:
As with https://github.com/TrianguloY/UrlChecker/issues/7 the main issue is to obtain the database with all the redirections to use.

Clear Urls have a public database with all the rules, that's the main reason it was implemented. In any case ClearUrl does some of the static redirections too (though probably not all).

stephenhawk8054:
As far as I see from Skip Redirect's issues, their approach is just using static redirections, and no need for a database.

I think most of the redirections are in this link:
[https://github.com/sblask/webextension-skip-redirect/blob/main/url.js](https://github.com/sblask/webextension-skip-redirect/blob/main/url.js)

I really don't know how to code in Java, so this is the best I can do to replicate the codes:

<details><summary>Java code</summary>

```java
import java.util.*;
import java.util.regex.*;
import java.io.UnsupportedEncodingException;
import java.net.URLDecoder;

public class main {
    public static String decodeURIComponent(String s, String charset) {
        if (s == null) {
            return null;
        }
        String result = null;
        try {
            result = URLDecoder.decode(s, charset);
        }
        // This exception should never occur.
        catch (UnsupportedEncodingException e) {
            result = s;
        }
        return result;
    }

    class Bundle {
        public LinkedList<String> prefix;
        public String prefixString;
    }
    
    public Bundle comprehend(LinkedList<String> pList, String s) {
        LinkedList<String> result = new LinkedList<String>();
        result.push("www\\.");
        for (String p : pList) {
            result.push(p + s);
        }

        String resultString = "(?:";
        for (int i = 0; i < result.size(); i++) {
            resultString = resultString + result.get(i) + "|";
        }
[...]

PR #110, opened 2022-09-30
Assets update action

PabloOQ:
This action updates assets files, currently there is only one asset file, the ClearURLs catalog, but  the action can be easily expanded if more are added ( maybe these issues can benefit from it: #82 #61 #10 #7 ), just copy and paste the ClearURLS catalog step and change the enviroment variables.

I think that even with the built-in catalog updater the app should come bundled with a recent file.

The script **needs** a hash url to confirm a successful download, I thought of removing this resctrition, but I don't like the idea of auto-commiting without some kind of verification. If a source doesn't have a hash to confirm it can be solved by forking the repo and adding an action to automatically sync with main every now and then, after that the action could just generate the hash.

The action won't fail if an asset could not be updated as it is aimed to update more than one file at the same time, these failures include:

- Hash does not match
- Folder does not exist
- The new file hash is the same as the current one
- Download failed

If the PR is merged it's important to keep in mind that the field `on:` is intentionally left out, I leave to you to decide when to run it the action, these are some that I thought could be useful:

- Manually

```
# manually
workflow_dispatch:
```

  - On release, this one would update the file after the release tag, so the file won't be technically updated until the next release
```
 # on releases
 release:
```

- I think actions can be called from other actions, so this can be added in an action that makes the releases (apk and all that) too .

- Weekly

```
 # weekly
 schedule:
   - cron: '0 0 * * Sun'
```

Finally, here are some [tests](https://github.com/PabloOQ/UrlChecker/actions). Because the action does not fail if the asset update fails you would need to enter to the workflow run and check the console log of the steps to see the result.

### DIFF
--- a/.github/scripts/file-updater.sh
+++ b/.github/scripts/file-updater.sh
@@ -1,0 +1,48 @@
+# Retrieve parameters
+fileName=$1
+fileURL=$2
+path=$3
+hashURL=$4
+
+# Temporal folder to work on
+# Needed in case a file is updated in the main folder, and the hash checking fails
+downloadFolder="./.file_updater_action_temporal_folder/"
+mkdir $downloadFolder
+tempPath="$downloadFolder$fileName"
+
+# Store checksum
+curlHashExit=1
+if [ "$path" != "" ]; then
+  checksum=$(curl $hashURL)
+  curlHashExit=$?
+fi
+
+# Avoid download of same file
+oldFileChecksum=$(sha256sum "$path$fileName")
+if [ "$oldFileChecksum" = "$checksum  $path$fileName" ]; then
+  echo "Current file checksum is the same as the new one"
+  exit 1
+fi
+
+# Download file
+curl "$fileURL" -o $tempPath
+curlFileExit=$?
+
+# If both downloads were succesful
+if [[ $curlFileExit -eq 0 && $curlHashExit -eq 0 ]]; then
+  # Checksum of file
+  dlFileChecksum=$(sha256sum $tempPath)
+  # If matches
+  if [ "$dlFileChecksum" = "$checksum  $tempPath" ]; then
+    mv $tempPath "$path$fileName"
+    # No need to check exit code, if mv fails GitHub actions will too
+  else
+    # If not, failure
+    echo "Hash does not match" >&2
+    exit 2
+  fi
+else
+  # If not, failure
+  echo "Downloads failed" >&2
+  exit 3
+fi
--- a/.github/workflows/assets-updater.yml
+++ b/.github/workflows/assets-updater.yml
@@ -1,0 +1,41 @@
+# This action updates assets files
+name: Assets updater
+
+on:
+  
+permissions:
+  contents: write
+
+jobs:
+  download:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository content
+        # https://github.com/actions/checkout
+        uses: actions/checkout@v3
+
+      - name: ClearURLs catalog
+        continue-on-error: true
+        run: |
+          bash ./.github/scripts/file-updater.sh $fileName $fileURL $filePath $hashURL
+          git add "$filePath$fileName"
+          echo "$fileName added successfully"
+        env:
+          fileName: "data.minify.json"
+          fileURL: "https://rules2.clearurls.xyz/data.minify.json"
+          filePath: "./app/src/main/assets/"
+          hashURL: "https://rules2.clearurls.xyz/rules.minify.hash"
+
+      - name: Commit everything and push
+        run: |
+          # Retrieve finished steps as number value
+          count=($(git diff --staged --name-only | wc))
+          if [[ ${count[0]} -ne 0 ]]; then
+            git config user.name GithubActions
+            git config user.email ""
+            git commit -m "Update assets"
+            git push
+          else
+            echo "None of the files were updated"
+          fi
--- a/.github/workflows/assets-updater.yml
+++ b/.github/workflows/assets-updater.yml
@@ -2,40 +2,59 @@
 name: Assets updater
 
 on:
-  
+ # weekly
+ schedule:
+   - cron: '0 0 * * Sun'
+    
 permissions:
   contents: write
+  pull-requests: write
 
 jobs:
   download:
     runs-on: ubuntu-latest
 
+    env:
+      workingBranch: assets-updater-action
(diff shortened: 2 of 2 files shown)


TrianguloY:
I must say I love this new action. It now seems obvious but I didn't think of it before! 


There is one thing that could be a bit dangerous: automatic commits.
Ideally the action should create a new branch (unless it already exists), make the commit there, and then create/update a PR for us developers to check and merge. This way we have more control over the changes (and we are notified of them!).

As for the trigger, with the manual merge, a cronjob is the best option (once a week for example). I'm not sure if an action can be triggered when an external repository changes, but it if were possible it would be the best (it will only run when the source repo changes!)


Do you want to try to modify it yourself? If not I'll definitely play with it.

TrianguloY:
In the meantime, I added you as a collaborator, feel free to join or ignore it.

If you decide to join just please continue creating PRs for normal and big changes. Small modifications or even bugfixes are ok, although I wouldn't mind a PR if you prefer.

PabloOQ:
> Ideally the action should create a new branch (unless it already exists), make the commit there, and then create/update a PR for us developers to check and merge.
Do you want to try to modify it yourself?

I can change it, shouldn't be too hard.

>  I'm not sure if an action can be triggered when an external repository changes, but it if were possible it would be the best (it will only run when the source repo changes!)

If this is possible, there should be a cooldown or a check timestamp to avoid more than one PR  on the same day, which I think should be more problematic. [On June 29, 9 commits were made on the ClearURLs repo](https://github.com/ClearURLs/Rules/commits/master) . Even then, each commit would trigger a workflow run. I don't think it is possible to make these checks before the workflow run. I will look into it but I don't expect much.

> In the meantime, I added you as a collaborator, feel free to join or ignore it.
If you decide to join just please continue creating PRs for normal and big changes.

Thanks. Will do!

TrianguloY:
> If this is possible, there should be a cooldown or a check timestamp to avoid more than one PR  on the same day, which I think should be more problematic. 

Ideally the script should reuse any active pr if it exists, by committing to the branch. This can be hard though, and as of now the weekly run is enough, so don't worry if you can't find how to do it.

The script could do something like this:

1) check if the custom branch exists, checkout it if it does, create if doesn't
2) Download the new file, then check the hash.
3) if the hash doesn't match, revert the file, if it does create a commit with the info and that change
4) repeat steps 2-3 with each file (currently only 1)
5) if there are new commits, push them
6) if something was pushed, and a pr does not exists, create one

Something like that may work, perhaps there is an easier way though

PabloOQ:
That should be it.
I swapped steps 4 and 5. Also if something fails in step 3 github will abort it, this will avoid the commit and push.

Step 6: I did try to check if there already was a PR, but it is not needed, a head can't have more than one PR pointing to the same base, which makes sense. Because of that the command used in the action fails (the step too) when trying to do a PR if there's already one, however the commits of the assets are pushed before this step so the PR will have them. This also makes the workflow be shown as failed in the actions tab.

The branch does not auto delete after the PR is closed, I think this can't be done here, but it shouldn't be a problem, as the only  files that are modified are the assets.

[Here is the action working.](https://github.com/PabloOQ/UrlChecker/pull/1) There are commits from this branch too because the workflow is executed on top of the changes made in this branch.

To make it work I had to change the settings of my fork `Settings->Actions->General` at the end there's an option to let actions create PRs.

Finally thanks to @Chgv99 who helped me test if the action could cascade into forks making them do PRs to `origin/master`. It does not.

TrianguloY:
> That should be it. I swapped steps 4 and 5. Also if something fails in step 3 github will abort it, this will avoid the commit and push.

Perfect, A commit for each file is even better.

> Step 6: I did try to check if there already was a PR, but it is not needed, a head can't have more than one PR pointing to the same base, which makes sense. Because of that the command used in the action fails (the step too) when trying to do a PR if there's already one, however the commits of the assets are pushed before this step so the PR will have them. This also makes the workflow be shown as failed in the actions tab.

It would be better to detect this situation to avoid failing a pr in this situation (so you can be sure that if the action failed it was for an unexpected reason), it's not really important though, it would probably never happen.

> The branch does not auto delete after the PR is closed, I think this can't be done here, but it shouldn't be a problem, as the only files that are modified are the assets.

That's something that the repository can be configured as, and I think it's set as such. When merging it should be auto-deleted. If not I guess the action will simple start from the old one, which as you said is not important since only the assets are updated. If really necessary I guess it can make a rebase first (which will do nothing if rebased already).
 
> To make it work I had to change the settings of my fork `Settings->Actions->General` at the end there's an option to let actions create PRs.

Already enabled :+1: 

> Finally thanks to @Chgv99 who helped me test if the action could cascade into forks making them do PRs to `origin/master`. It does not.

Thanks both of you for the extra check!

---

I have just one more question: why do you use a temporal folder instead of simply downloading it directly? It runs under git, so you should be able to revert it to the latest commit version on error. Not really important, just wondering.

PabloOQ:
> It would be better to detect this situation to avoid failing a pr in this situation (so you can be sure that if the action failed it was for an unexpected reason), it's not really important though, it would probably never happen.

I did try to find if there was a PR already, it should be possible with `gh pr list` , but I couldn't make it work. I am probably misunderstanding the sintaxis of `--base` and `--head`. Still, to the best of my knowledge I tried every combination ( `origin`, `main`, `master`, `assets-update-action`, `origin/assets-update-action` , `origin:assets-update-action` ,... ) and it wouldn't work. And with the [search query syntax (`--search`)](https://docs.github.com/en/search-github/searching-on-github/searching-issues-and-pull-requests#search-by-branch-name) if I searched for `assets-update-action` it would detect `assets-update-action-2` too, which will probably never happen, but I rather not do the check than do it wrong. Because this would only fix how it's shown in the actions tab I thought it wasn't that bad. Feel free to give it a shot if you want.

> I have just one more question: why do you use a temporal folder instead of simply downloading it directly?

Honestly I didn't thought of reverting changes. At first I was using [git auto commit](https://github.com/stefanzweifel/git-auto-commit-action), which adds and commits everything, except hidden folders if I remember correctly. I was also downloading things on the root folder and if the hash failed I would remove them, if the hash matched I would move them to their respectives folder. With that in mind if an asset path was the root folder itself and the hash failed it would remove the file and then commit it. To solve that I would just download things in the hidden folder.

I decided to ditch git auto commit because adding everything is a bad praxis, also I wanted to avoid using actions that weren't from github.

PR #112, opened 2022-10-01
Ilithy patch 1

Ilithy:


TrianguloY:
👀

Btw, feel free to create new labels if you want (you can, don't you?). The "ongoing translation" for example was a one-time label that I should probably remove, just "translation" could be a new one.

PR #113, opened 2022-10-01
Basic PR labeler

Ilithy:
Small proposal, after a lot of testing (of nonsense too) and back and forth, here is a "basic" labeler (it will surely require additions and modifications).

It labels the PRs of:

- modifications of language files (strings.xml): adds the double label "Ongoing translation" and the language code (for example "es-ES" for Spanish) according to the modified file.

- internal files ("Core" label)
when editing files in the `app/src/main/java/com/trianguloy/urlchecker` folder and all its subfolders

- Adds the "Github" label when modifying files in the `.github` folder and subfolders

- Adds the label "Assets" when modifying files in the `app/src/main/assets` folder and its subfolders

- Adds the label "Store file" when modifying files in the `app/src/main/play` folder and its subfolders

- Adds the label "Gradle wrapper" when modifying files in the `gradle` folder and its subfolders

It works (normally) perfectly well.
(Sorry for the delay)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -12,7 +12,8 @@ Github:
 
 # Add 'Core' label to any change within the 'core' package
 Core:
-- app/src/main/java/com/trianguloy/urlchecker/**/*
+- app/src/main/java/**/*
+- app/src/main/kotlin/**/*
 
 # Add 'Store file' label to any change within the 'play' folder and subfolder(s)
 Store file:
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -22,14 +22,10 @@ Store file:
 Gradle wrapper:
 - gradle/wrapper/**/*
 
-# Add 'Ongoing translation' label to any change to strings.xml files
+# Add 'Ongoing translation` label to any change to strings.xml files as long as the English strings.xml file hasn't changed
 Ongoing translation:
-- app/src/main/res/values-es/strings.xml
-- app/src/main/res/values-fr-rFR/strings.xml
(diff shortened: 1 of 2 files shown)


TrianguloY:
Looks nice and useful!

A couple comments:
- Use the label 'Translation' instead of 'Ongoing translation' (I already edited it)
- Don't tag individual locales, currently there are just a few, but theoretically they could be hundreds, it's just not necessary, 'Translation' is enough

.github/labeler.yml
```diff
@@ -0,0 +1,48 @@
...
+Assets:
+- app/src/main/assets/**/*
+
+# Add 'Github' label to any git file changes within the source dir
+Github:
+- .github/**/*
+
+# Add 'Core' label to any change within the 'core' package
+Core:
+- app/src/main/java/com/trianguloy/urlchecker/**/*
```
TrianguloY, line 15:
```suggestion
- app/src/main/java/**/*
- app/src/main/kotlin/**/*
```

although it would be strange to add files outside the main package, they are indeed possible. the 'java' (and 'kotlin', currently unused but maybe in the future it is) folders are where all the 'core' code must be added

Ilithy:
The changes have been made,
too bad for the language code, it looked good ;-)

.github/labeler.yml
Ilithy: Thanks for the clarifications, and the corrections, they have been added

TrianguloY:
yeah, sorry to 'destroy' your work, but it would be really hard to maintain in the future :(

merging!

PR #120, opened 2022-10-09
Copy to clipboard from debug module

PabloOQ:
I made it because I found an app that wouldn't open links from URLCheck but it would from [Open link with](https://github.com/tasomaniac/OpenLinkWith/) after being shared from URLCheck (did not open either when sharing from URLCheck to URLCheck). When I wanted to compare the Intents ( `App->URLCheck-share->Open link with->URLCheck` ,  `App->URLCheck` ), I had to take 2 screenshots and I thought it would be more convenient to be able to copy it to the clipboard. I will open an issue or fix it after doing more research.

When tapped it will copy the `text` from the `TextView` that has been tapped, when held it will copy both.
Because this module is for developing I think the strings should match the code, that's why I didn't translate `Intent` and `UrlData`,

The second commit is to avoid code duplication.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,7 +7,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.3.0'
+        classpath 'com.android.tools.build:gradle:7.2.2'
         
 
         // NOTE: Do not place your application dependencies here; they belong
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/DebugModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/DebugModule.java
@@ -1,13 +1,8 @@
 package com.trianguloy.urlchecker.modules.list;
 
-import android.content.ClipData;
-import android.content.ClipboardManager;
-import android.content.Context;
-import android.os.Build;
 import android.view.View;
 import android.widget.CheckBox;
 import android.widget.TextView;
-import android.widget.Toast;
 
 import com.trianguloy.urlchecker.R;
 import com.trianguloy.urlchecker.activities.ConfigActivity;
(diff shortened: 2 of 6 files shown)


TrianguloY:
The debug module was made for myself as developer (originally hidden, later public just in case someone found it useful). Any functionality you can think of that can help us is welcomed! 

This copy feature is nice, I like it. Another alternative would have been to set the textview as selectable like an editbox (but without being able to edit it, so you could long tap and then move the text handlers to choose what to copy). I think that would have been easier to implement, in any case your alternative is useful too.

As for the toast, I would have reuse a general "copy to clipboard" one, easier to translate too. Think that maybe in the future we add more elements, having a dedicated toast for each one can become cumbersome.

In any case, as I said this module is for us developers, so I don't really mind the changes there. Modify them or merge them, as you prefer. (Except for the downgrade comment, that is necessary to be fixed first)

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.3.0'
+        classpath 'com.android.tools.build:gradle:7.2.2'
```
TrianguloY, line 10: Careful with this downgrade, maybe a bad merge? Remove this change, please.

TrianguloY (changes requested):

build.gradle
PabloOQ: Don't know how that got there, is fixed now

PabloOQ:
> As for the toast, I would have reuse a general "copy to clipboard" one, easier to translate too. Think that maybe in the future we add more elements, having a dedicated toast for each one can become cumbersome.

Sometimes the debug module is too big (mainly urls made for tracking, such as redirections, and queries) and I felt some feedback would be nice, to see if I copied the part I wanted to copy.

PR #128, opened 2022-10-21
Update assets-updater.yml

PabloOQ:
Now auto-deletes the working branch when no assets are updated.
Now creates PR only when there's none.
As a side effect, it _always_ creates the branch on top of `master`. In practice it doesn't change anything, in theory [what I mention here (the paragraph with the hyperlink)](https://github.com/TrianguloY/UrlChecker/pull/110#issuecomment-1267676725) wouldn't happen again.
I thought of doing a commit directly but I don't like submitting actions without it being reviewed first.
The [message of the PR wasn't displaying properly](https://github.com/PabloOQ/UrlChecker/pull/4) (I think it is related to the indent of the new if statement), so I moved it to the env variables.

[Some tests.](https://github.com/PabloOQ/UrlChecker/actions?query=branch%3Atest-update-action-3)

### DIFF
--- a/.github/workflows/assets-updater.yml
+++ b/.github/workflows/assets-updater.yml
@@ -21,8 +21,11 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Checkout main
+      - name: Checkout master
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+          ref: master
 
       - name: Git config
         run: |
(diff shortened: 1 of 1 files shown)


TrianguloY:
It's incredible how complex and advanced git is, yet something as simple as 'add this changes, create or update a pr with them' requires so much checks. In any case this was needed, thanks!

I wonder if using a different step for each file would simplify things, having to check just 1 file on the pr may be easier (if it fails, no need to continue). A PR with each file change is not bad either, maybe even preferred. Things to consider, but not really needed.

PR #129, opened 2022-10-24
updated assets-updater.yml action

TrianguloY:
- each possible file is now a separate job (different branch and PR): I think it makes more sense this way, and simplifies a lot all the logic
- Uses the branch as base if exists: ideally we should not touch this branch, but if we do the action will use it. This also means that if a PR is made but the branch is not deleted it will use it as base. This is not bad though, and we can rebase it if necessary.
- updates only the necessary: it doesn't create+delete a branch if there are no changes for example
- no temporal directory: the file is directly downloaded in the final destination, comparisons are made with the hash (but git could be used if in the future we have files without hash)
- inline script: I understand the need to have the script in a separate file, but I also like to have related info together. I'm still not sure what is better

@PabloOQ since you made most of the previous versions I request your review in case you find a problem or something that could be improved, if you have time and want to check. I made some tests on my private repo so unfortunately I can't show them, but you can copy it to yours and made some there if you want.


PabloOQ:
I did not find any bugs and everything is straight forward. Looks good to me. I took the liberty of adding a label ( `action` ) to the automatic PR, also I already went back to previous PRs (#118, #119) and labeled those manually.

PabloOQ (approved):

PabloOQ:
Weird, I only merged the branch, don't know why it deletes it too. Anyways is still up in case you want to check something.

TrianguloY:
> I did not find any bugs and everything is straight forward. Looks good to me. I took the liberty of adding a label ( `action` ) to the automatic PR, also I already went back to previous PRs (#118, #119) and labeled those manually.

Nice addition! Thanks for reviewing it.

> Weird, I only merged the branch, don't know why it deletes it too. Anyways is still up in case you want to check something.

I enabled it, it deletes the branch automatically but as you saw it can be restored if needed. I deleted it again (no need to have outdated branches)

PR #132, opened 2022-10-26
New badge

Ilithy:
I take the liberty of proposing new (minimal) changes:

* Added some small badges (I'm fighting my addiction to badges, but it's hard)
* Slight modification on the header image (to make it a little smaller in height)

I leave you judge, and executioner ;-)


TrianguloY:
I'm sorry, but I'm going to decline this PR :(

Here are the reasons:

- GitHub last commit: this information can be seen directly on the main page at the top (just above the code files). As I have explained in the past I prefer to avoid adding duplicate information, specially with badges where having too many is very chaotic, or at least I find it confusing (I've seen repositories with three lines of badges. They look colorful but that's it). Probably you saw the new badge I added yesterday, the commits/month. This information is not available on the main page, and the users can see if the repo has activity or not, that's why I added it.
- workflows status: This can also be seen on the main page (in the same place, although it's just a common green tick/red cross that you must click) but I personally don't like it for the following reason: It can confuse users into thinking that the store app is 'wrong' (at least that's what I though in the past with other apps). Whether the workflows passed or not on the last commit is useful for us developers, but not for end users. I prefer to keep the Readme as something for everyone, where possible.

Having said this, I still like all the little badges with the info. I'm thinking of placing all these, and more, either on a separate page or in a collapsible element. I'm open to ideas, and of course my decisions are debatable.



- The featured image is unfortunately impossible to change. Play Store requires that image to be exactly 1024x500. [I should add this (and other image related checks) to the triple-t check action.]

But, it is possible to have a separate image for the GitHub readme, and I think I prefer it (it is too big, I agree). It should be a copy (we'll have a bit of a duplicate there, but I think it's worth it) and, to avoid even more files on the root, we can place it in a /repository/ folder. Feel free to create a PR with this change.

Ilithy:
>I'm sorry, but I'm going to decline this PR :(

No problem, I understand, thanks for the clear explanations, these are good points.

>The featured image is unfortunately impossible to change. Play Store requires that image to be exactly 1024x500. [I should add this (and other image related checks) to the triple-t check action.]

Clear, I understand. For the addition of the images to the Triple-t verification, it is an excellent idea 👍

>But, it is possible to have a separate image for the GitHub readme, and I think I prefer it (it is too big, I agree). It should be a copy (we'll have a bit of a duplicate there, but I think it's worth it) and, to avoid even more files on the root, we can place it in a /repository/ folder. Feel free to create a PR with this change.

I can store it on my repo (Ilithy/Ilithy) so as not to weigh down this repo unnecessarily, as you wish

PR #155, opened 2022-12-12
[Proposal] Readme changes

Ilithy:
Notable (proposed) changes:
- added two new captures to the readme (7-modules and 8-settings.png)
- moved contributors to a dedicated section
- change in the display of screenshots, for more flexibility in the display resolution of the readme
- added details in the presentation text
- some other slight modifications...


TrianguloY:
I'm still not sure about the new header, it is _too big_, maybe on light theme looks better, but on dark theme it looks like this:
image

Hmm, what about without text?
image


And even though the app name is there, I prefer to have it as text because some users have images disabled, and also because the human mind usually ignores big image headers ([it's true!](https://en.wikipedia.org/wiki/Banner_blindness))

The rest of the proposed changes are really nice, thanks!

Once we agree on the header I'll merge the changes.

Ilithy:
Your new header is very well thought out, I find it perfect 👏

Ilithy:
Thanks for restoring the title for me 🙏

PR #156, opened 2022-12-13
Update Pattern Checker description

Ilithy:
Added a little more info on Pattern Checker module description.
if it doesnt fit for you, ignore this.

Thanks.


TrianguloY:
Hmm, it's true that it needs a better explanation, but even though yours is good I think it could be improved even more. What about:

```
This module checks the url with regexp patterns that warns, suggests or applies replacements. You can modify or create your own patterns, or even use user-created ones.
Built-in patterns include:
* Warning when contains non-ascii characters like greek letters. This can be used for phishing: googĺe.com vs google.com
* Suggest replacing 'http' with 'https'
* Suggest replacing Youtube, Reddit or Twitter with privacy-friendly alternatives [disabled by default]```

The full-description could be improved too in a similar way...

Ilithy:
> Hmm, it's true that it needs a better explanation, but even though yours is good I think it could be improved even more. What about:
> 
> ```
> This module checks the url with regexp patterns that warns, suggests or applies replacements. You can modify or create your own patterns, or even use user-created ones.
> Built-in patterns include:
> * Warning when contains non-ascii characters like greek letters. This can be used for phishing: googĺe.com vs google.com
> * Suggest replacing 'http' with 'https'
> * Suggest replacing Youtube, Reddit or Twitter with privacy-friendly alternatives [disabled by default]
> ```
> 
> The full-description could be improved too in a similar way...

Yes, your version is much more complete and I find it explicit enough, I apply this updated description on this PR 👍

TrianguloY:
> warn for missing http(s) scheme

I opted to omit this because it's just a pattern to avoid having a 'no app available' when the link doesn't starts with http. It's something the app needs, not something the url needs, and I think it could confuse more.

Anyway, I fixed the indentation and added the spanish translation. You forgot the french strings! And I guess you will want to update it, so once you upload the file tell me and I'll 'squash and merge' this pr (or you can do it yourself)

Ilithy:
>I opted to omit this because it's just a pattern to avoid having a 'no app available' when the link doesn't starts with http. It's something the app needs, not something the url needs, and I think it could confuse more.

It suits me, I followed your model for the French version.

>Anyway, I fixed the indentation and added the spanish translation.

Thanks

> You forgot the french strings! And I guess you will want to update it,

Yes I skipped this step --', thank you for the reminder, it's done 👍

>so once you upload the file tell me and I'll 'squash and merge' this pr (or you can do it yourself)

I let you manage this 🙂

Thanks

PR #157, opened 2022-12-17
Added italian translation files

dperruso:
Added italian translation files in the correct folders. I'll do any corrections after using the app in italian, to better understand context.


TrianguloY:
Thanks!

Here is the app with your locale added.
[app-alpha.zip](https://github.com/TrianguloY/UrlChecker/files/10254316/app-alpha.zip)

It is an alpha version, it will NOT remove the standard version and you will have both apps at the same time. Please test it and make all the changes you want, once you are happy just tell me and I'll merge it.
Of course you can update the locale in the future whenever you desire.

dperruso:
Whoops, sorry for the double commit. 
It should be all good now! 
Also, thanks a lot for providing me with an apk to test my translations! :)

TrianguloY:
Don't worry, I'll squash all commits in a single commit.

Ah, you already fixed the double quotes. They are needed because the ' is not valid. An alternative is to escape it with \' but personally I prefer the double quotes.

I'll merge it tomorrow

PR #159, opened 2022-12-19
[Action] Release apk on GitHub

Ilithy:
Hi;
This PR introduces a github action to create an apk file and make it available on github.
- the action is automatic when creating a new version (when a new tag is created)
- the action can be launched manually

Other changes:

- Added 'get it on github' badge
- Added 'get it on APKPure' badge
- Modification of the size of the "_get it on_" badges (to have them on one line on a majority of screens)

Thank you.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,32 @@
+name: Build & Publish Release APK
+
+on:
+  push:
+    tags:
+      - '*'
+
+  workflow_dispatch:
+
+jobs:
+  Gradle:
+    runs-on: ubuntu-latest
+    steps:
+    - name: checkout code
+      uses: actions/checkout@v3
+    - name: setup jdk
+      uses: actions/setup-java@v1
+      with:
+        java-version: 11
+    - name: Make Gradle executable
+      run: chmod +x ./gradlew
+    - name: Build Release APK
+      run: ./gradlew assembleRelease
+    - name: Release to GitHub
+      uses: Ilithy/upload-release-action@v2
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        file: ./app/build/outputs/apk/release/app-release-unsigned.apk
+        asset_name: URLCheck.apk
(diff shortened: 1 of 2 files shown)


TrianguloY:
I appreciate the suggestion and your work creating the action, but I'm afraid its not so easy to create an apk (well, it's not hard either, but requires a few more lines and effort).

An android apk is basically a zip with sort of compiled code and assets, and a few other information about the app it contains. One of this information is a certificate that '[signs](https://developer.android.com/studio/publish/app-signing)' the apk and ensures that only the original author was the creator of it. This is a privacy measure to ensure that malicious external apps/users can't replace your F-Droid apk with their copy, for example. An Android device will only allow updates for an app if it was signed with the same certificate as the installed version.

All apks must be signed before being installed, and since while you are developing you don't mind which one to use, the android gradle framework automatically uses a 'debug' certificate for all debug builds. This certificate is public and available anywhere, so you shouldn't build a production apk with it.

On Play Store I use my own certificate, on F-droid they use their own. This is also the reason why you can't usually update from play store an app that was installed from f-droid, and vice-versa (although f-droid supports providing your own certificate somehow...but I haven't analyzed that).

The alpha app, and other apks created with actions without any signing steps, are not. They are fine for being alpha, but not if it were to be published as 'final'. Your PR is one of those, so unfortunately I cannot merge it.

A solution would be to upload my certificate as a secret here on github, and use a step to sign the built apk, but I would prefer not to do that if possible. Another solution would be for me to build it locally (which I already do for Play Store) and upload it manually.

The main thing is that f-droid already allows anyone to download their signed apk, so even though I understand that having the apk directly here on this github would be convenient, I'm not really sure if the extra effort required is that important.

In any case I'll keep this PR open to see your opinion on this.

<hr>

Oh, and about that other extra badge, the thing is that I'm not really fond of those service that extracts files and publish them themselves on their (usually ad-filled) pages. My license is to mention me, and they do while copy/pasting the description so it's ok, but still they don't deserve me linking them. (F-droid is different, as I choose my app to be published there)

Issue #32, opened 2022-07-17
Open in app configuration setup wizard/tutorial


TrianguloY:
On Android, apps can be configured to open automatically an url matching some characteristics. This app is configured to match all urls, but some are not opened because other apps have more preference, which is not easy to change specially for non-advanced users.

Add a screen where you can paste an url, and it will tell you the app that is currently configured to automatically open that link, and if it's not ours include a guide/steps to how to configure it, if possible (this may be difficult because each android version/manufacturer has a different UI, consider adding links to external documentation). 
If there were a way to automatically configure it by pressing a button that would be awesome.
I've seen a similar dialogs on browsers (firefox to be precise), where you can press a 'set as default' button and a dialog with all currently configured browsers is opened for you to choose directly., to be considered.

Ilithy:
>If there were a way to automatically configure it by pressing a button that would be awesome.
I've seen a similar dialogs on browsers (firefox to be precise), where you can press a 'set as default' button and a dialog with all currently configured browsers is opened for you to choose directly., to be considered.

I don't know if this will be useful, but there is something similar in the [link eye](https://github.com/kuesji/link_eye) application (written in Java), you press a button, and you are taken to the default application selection settings:

![Screenshot_20220817-174728](https://user-images.githubusercontent.com/36798218/185185401-cf77681c-e00c-45be-a9f5-2bc243716cd0.png)
**_After clicking on Enable, we are redirected to the default app page_**

![Screenshot_20220817-174734](https://user-images.githubusercontent.com/36798218/185185489-0713221c-73ed-441d-acf6-39f4c8802916.png)
_Sorry the capture of my settings is in French_

TrianguloY:
Hmm, that's a good start, and probably worth implementing as a first step for general links (I think that dialog is not present in all android versions, but must be on most of them).

With this issue however I was thinking of specific links that are set to be opened on specific apps, like twitter, github, google maps, etc. To change those you need to go to the configured app (the one that opens them) and change it there, which will require a dialog to test different links, either by the user or some common ones. It's not difficult I think, but I'm not sure yet how to do it useful for the user (for this reason I added the 'help wanted' tag)

Ilithy:
>Hmm, that's a good start, and probably worth implementing as a first step for general links (I think that dialog is not present in all android versions, but must be on most of them).  

_If I'm not mistaken, the default application selection has been present since Android 9 (api level 28)_
[...]

PR #164, opened 2023-01-11
Tutorial screen

PabloOQ:
Happy new year!

This PR is more of a setup for the programming part of the feature than a fully functional tutorial. I made the different screens but they are a bit ugly and need some changing, also I expect the wording of the tutorial to be changed. That's why there is no spanish translation, if the PR goes through we could open a discussion or a issue and talk about it there, after that I'll do the translation myself.

It doesn't seem this is what #32 is about, anyways I still think this could be useful.
I have been wanting to implement this for a while, I did want to try to make the wizard setup with the browser prompts too, but I have been busy for quite a while. Finally found some time and decided to do a tutorial for the app as it is not clear enough how the app works or how to setup it properly, or at least that's the feedback I got after recommending the app to a friend who is tech savy enough to understand these things easily.

I hope everything is correct, this is the third iteration. I tried implementing it with `ViewPager`, which uses `androidx`, I don't know why but I decided to check the app size after the implementation was done, and not when adding the libraries, that was a mistake, the apk was 4 times bigger! Anyways I decided to do it again, as I think you wanted to keep the app size as small as possible. The second time I did the change of views manually, I didn't like it. Finally decided to go with `ViewFlipper` and I think it looks good.

I wanted to reuse the buttons from the wizard/settings so I moved them to a `Fragment` as they need to be able to have `onActivityResult`, the problem is `Fragment` is deprecated, you even made a `Fragment` interface, but it's only for the modules config. Because of this I did the browser buttons extraction in a different commit, in this commit the buttons are extracted from the settings into a `Fragment` which is included in the XML of the settings activity and the `View` of the tutorial which explains the buttons. So feel free to just revert it.

Let me know what you think!


TrianguloY:
Wow! This is not what #32 was about (that was for a way to enter a url and help configuring it so that it is opened in this app if it isn't). But this tutorial is a great addition! I'm impressed, good job.

I glanced over the code and tested it a bit myself (personal tip: when you create a PR on a repository which includes UI changes, like this one, add some pictures to show those changes. That help us reviewers a lot!) and even though I think the code will probably need to be refactored a bit, and the texts can be slightly improved, the idea and general functionality is solid and I'm sure very useful to new users.

> I hope everything is correct, this is the third iteration. I tried implementing it with ViewPager, which uses androidx, I don't know why but I decided to check the app size after the implementation was done, and not when adding the libraries, that was a mistake, the apk was 4 times bigger! Anyways I decided to do it again, as I think you wanted to keep the app size as small as possible. The second time I did the change of views manually, I didn't like it. Finally decided to go with ViewFlipper and I think it looks good.

I'm aware of that. Keeping the app as small as possible is something I personally want to do for personal reasons (my first phone was a very low-end device and hated when apps used all the memory just to show a useless animation, so I decided to make my apps really simple and just what was needed, and so far I have kept it). This of course comes with a price, not being able to use some more 'newer' solutions like androidx or even kotlin. I'm open to switch and use them if they really add value and give something that is almost impossible to do by myself or with alternatives, but right now that's not the case. That ViewFlipper looks good enough, I have never use it myself but seems very easy.

> I wanted to reuse the buttons from the wizard/settings so I moved them to a Fragment as they need to be able to have onActivityResult, the problem is Fragment is deprecated, you even made a Fragment interface, but it's only for the modules config. Because of this I did the browser buttons extraction in a different commit, in this commit the buttons are extracted from the settings into a Fragment which is included in the XML of the settings activity and the View of the tutorial which explains the buttons. So feel free to just revert it.

Yup, Fragments were a nice addition but for some reason they deprecated it. The alternative is that huge compatibility library but...in the end they are just a small activity that can be included in other activities. If the only requirement is the onActivityResult listener, we can either hardcode it or create our own hook. I'll think about this.

In any case give me a few days to thoroughly review this changes, because I really like this new feature and I want to check if I can improve those points or leave them as-is. Later, as you said, we can improve the texts and provide translations.

Thanks again!

TrianguloY:
Btw, for other watchers of this PR, here are pictures of this new tutorial:

Screenshot_1673480010
Screenshot_1673480012
Screenshot_1673480015
Screenshot_1673480017

PabloOQ:
> This of course comes with a price, not being able to use some more 'newer' solutions like androidx or even kotlin.

Well I've got to say that if the app wasn´t written in java I probably wouldn´t have made the first PR, as I don't feel comfortable in kotlin. I was specifically looking for an app with ClearURLs or some kind of url interceptor, and then write the code for ClearURLs myself if it wasn't there. To my surprise after weeks of trying to find an app with these characteristics I stumbled accross URLChecker and it was better than what I was trying to find. The modules are just great for customizaton and make starting to write code easy.

> personal tip: when you create a PR on a repository which includes UI changes, like this one, add some pictures to show those changes. That help us reviewers a lot!

> Btw, for other watchers of this PR, here are pictures of this new tutorial

Thanks for uploading the images, I honestly didn't think about it, I'll keep it in mind!

TrianguloY:
Update: as you have probably notice I've been out of programming for quite some time now (in fact I haven't even turned on my computer since last week I think).

Next weekend I'll finish this "vacations" and continue working on this and other features!

TrianguloY:
I've tweaked a few things, removed the fragments dependency (with a custom injector which...is not pretty but works reasonably well) and changed a few things from the screens. See below!

   

The second screen is now very crowded, I think it can be separated in two (first paragraph / rest). What do you think?

PabloOQ:
> The second screen is now very crowded, I think it can be separated in two (first paragraph / rest).

Yes, I agree, that seems like too much text for one screen.

About the buttons to skip/end the tutorial. I did not combine them on purpose with the navigation buttons to avoid skipping the tutorial if one were to double tap by accident on the penultimate screen. I guess it is ok to combine them as this was way too specific.

Aside from that I don´t see anything weird, as soon as it is merged I'll start working on the spanish translation.

TrianguloY:
Hmm, that's a very valid concern, and I think it can be avoided by placing a "double click" check: if the next button is pressed again less than a second from before, ignore. Maybe the same with the previous button, but in that case the device back button should not be restricted (just in case someone want to quickly exit)

TrianguloY:
Now the second page was quite empty: I don't quite like this 'diagram' but all the elements are dynamic so it would be translated and adapted to your device... any comments?

image image

PabloOQ:
I like the diagram of the first screenshot, nowadays icons tend to be more simple so it is more "consistant" with the current trend, the android 3D model makes it feel "old" to me. I thought of making a diagram but didn't know how to approach it as I don't know if what I had in mind could be translated, it was my main concern too. I couldn't find anything about making arrows with dynamic elements, which seems the only way to make the diagram easily translatable.
Here is the mockup, I like yours more anyway, it's more compact.
image

TrianguloY:
Funny you say the right one is an 'old' pic, because it's from an android 4.0.3!

Your diagram gave me an idea
    

Too much? I like it to be honest

Ilithy:
@TrianguloY Thank you, it's done, however I don't understand the error, the target element is present.

TrianguloY:
Thank you!

The error was because you had a duplicated end tag (which is far easier to see if you have an IDE that shows exactly where the error is 😅) it is fixed now

PabloOQ:
The android icon is too tiny in spanish. I made some changes to the layout (changed weights, removed padding). To fix it, still it doesn't look as good.

First screenshot is before the change, the others are after the change.

TrianguloY:
I tried to tweak it a bit for extreme resolutions, but nothing seem to work for everything so I just leaved it as it is now, will modify later if needed. I do detected some issues with scrolling, they are fixed now (as well as modifying the padding as it is now on the main branch).

I think there is nothing else to do, I'll probably merge it tomorrow (will join/amend a couple commits before the merge)

TrianguloY:
I couldn't make GitHub detect the manual merge, so I needed to force push the changes to your branch. I still keep a copy of the original commits just in case.

Anyways, it's merged now!

Issue #166, opened 2023-01-23
Scroll Bar is overlaping the content


2002suraj456:
**Describe the bug**
Content is overlapping the scrollbars as provided in the screenshot.

**Screenshots**

<img src="https://user-images.githubusercontent.com/49604350/214106014-22686687-da29-4393-9dfc-0bb4dfd295b2.png"  height="550">

PR #168, opened 2023-01-24
fix: Adjusted padding in the layouts. #166

2002suraj456:
removed the padding from the scroll view and
added them to the linear layout.

Also changed the padding in other activities where padding is done
in the scroll view.


2002suraj456:
Here is the screenshot.




I want to add some UI improvements to the app so should I discuss them in a single issue or have them in different ones?

TrianguloY:
that 1px is hard to get, and I'm sure that in other android versions/roms the scrollbar is draw differently so, the easier the better.

Merged!

PR #174, opened 2023-02-06
Update issue templates

Ilithy:
Hi;
A small PR to update the issue templates, switching them from markdown to Yaml syntax, for more flexibility.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms


TrianguloY:
Thank you! These new form-like issues are much better than manually filling a template 👍 

I made a few edits, mainly to fix a few grammar and clarify a couple inputs.

But I also noticed something! The 'ask a question' shouldn't be a template for an issue, but a discussion. Now that we have the discussions tab, those type of questions are better there (easier to reply, search and mark as answer too). Can you think of a way to redirect users that want to ask (or show) things to create a discussion instead? Maybe a direct link? (is it possible?)

Ilithy:
You are right, I kept it like this because there are always users who use it to piss off questions, but I will delete this model

>Can you think of a way to redirect users that want to ask (or show) things to create a discussion instead? Maybe a direct link? (is it possible?)

This could be added to the mandatory information at the bottom of the templates, like this _(this is just an example)_

TrianguloY:
Turns out there is a way to add links!
For some reason I can't see the issues link on your fork, in any case I'm going to merge it, can be changed later.

Thanks again for the change! (Switching from the old template to this new format sure required work!)

TrianguloY:
looks good!

Btw, You forgot to make the changes on a new branch, and I used a squash, so now both forks have diverged. I updated your fork to avoid future problems, the old commits are in https://github.com/Ilithy/UrlChecker/commits/update-issue-template (you can delete that branch if you want)

PabloOQ:
Just saw this for the first time in the issue I opened. The template looks great, really cool!

Ilithy:
Thanks @PabloOQ 🥰

PR #178, opened 2023-02-13
Update fr-FR strings

Ilithy:
Commit addition tracking: 6f15783

Translation of new elements into French


TrianguloY:
I made the same mistake 😅.
From the code I'm replacing those placeholders with integers, which requires %d. %s is for strings. There are other formats too (it's a standard)

PR #185, opened 2023-02-20
Add Simplified Chinese translation

Sevichecc:


Ilithy:
Hello, you didn't have to close this PR, the error came from the name of the folder containing the `strings.xml` file, renaming it `values-zh-CN` would have resolved the build error

Sevichecc:
> Hello, you didn't have to close this PR, the error came from the name of the folder containing the `strings.xml` file, renaming it `values-zh-CN` would have resolved the build error

Thanks for the reminder! I have also noticed some omissions in the translation, so perhaps I should submit another PR.

PR #186, opened 2023-02-20
Add Simplified Chinese translation

Sevichecc:


TrianguloY:
Apparently the name of the folder needs to have an extra 'r' (Android is very strange with locales)

Let's see if it works now...

Ilithy:
Sorry for the wrong information, indeed @TrianguloY is right, the folder name should be `values-zh-rCN`

TrianguloY:
Fixed!

Let me review it a bit and I'll merge it (just to make sure the translation is more or less expected and there is no hidden messages).

Thanks!

Sevichecc:
Sorry for the extra command, I use VS Code (but with Neovim extendsion 🤣

TrianguloY:
Apologies for the delay. Everything seems fine. Merged!

PR #193, opened 2023-02-24
Update PRIVACY POLICY

Ilithy:
Minor changes:

* Add information about url unshortener module.
* Displacement of the note on the Log module (doesn't have to be in 3rd party cloud dependencies)

Thanks


TrianguloY:
I think I'm going to disable the code review action, at least until we discover how to ignore some specific checks. Too many false positives and nothing worth keeping, I think.

Other than that, thank you! To be honest all your work with these non-programming tasks is really appreciated 😊
Merged!

Ilithy:
Yes indeed it seems to me reasonable, on this PR I realized clearly the useless points that it to raise.

It is with pleasure.

PR #198, opened 2023-03-03
New Flags Editor module

PabloOQ:
Solves #37.
Lots to unpack here, hopefully I don't forget to mention anything.
The Flags Editor module lets the user edit the flags when opening an URL from URLCheck.
The module dialog has 3 components:

- Edit field: Allows to directly edit the flags. It is not enabled, to enable it press the edit button, now you can edit the flags, press edit again to save these flags. Hold edit to change the flags with the ones the flags from the intent that launched URLCheck.
- Add flags: Lets you "search" (it's an autofill it doesn't actually search) for a flag an adds it to the intent.
- List of current flags: Can be shown by pressing the arrow on the right. Shows all the current flags that are in the list, because the module is heavily inspired by the Interceptor of app manager, [the list of flags is taken from there](https://github.com/MuntashirAkon/AppManager/blob/19782da4c8556c817ba5795554a1cc21f38af13a/app/src/main/java/io/github/muntashirakon/AppManager/intercept/ActivityInterceptor.java#L92). The flags are first added to a list and then checked through reflection, maybe unnecesary, but I think it simplifies maintenance, and lets us not worry about which flags are from which android API, downside is that we can't hardcode some special flags, like `FLAG_IGNORE_EPHEMERAL`.

Sadly I couldn't use `URLData` to pass the flags, as any module can create a new object and the flags are lost. So I made a first implementation with `GlobalData`, at first I made a singleton-like class to hold the data (first commit), but it had a bug, if you have the option from #109 disabled, you could share the intent to URLChecker itself and every instance would be pointing to the same `GlobalData`. Because of that I changed the implementation to one that I honestly don't like that much. In this second iteration the `MainDialog` implements an interface to keep its own `GlobalData`, modules need to cast the activity first to access this `GlobalData`, which is now one `GlobalData` per dialog. At the end I noticed, that I didn't format the strings correctly so the fix is in a separate commit. Because of that the history module can't keep track of the flags changed.

To show the flags as hexadecimal in the debug module they are stored as a 0x string, so I added the option to pre-process and post-process strings attached to an `EditText`, in the `GenericPref` file, as you can set default flags to substitute the ones that we get from the intent. I really like this one, as it allows for filtering anything from an `EditText` before storing it in disk.

To reuse code to tint buttons I extracted it from `HistoryModule` to the new `DrawableButtonUtils`. I reused some code from the Remove queries module too, the part that adds buttons for each query/flag.

There are 2 FIXME tags in `FlagsModule`, as it took longer than expected and I don't think I can revisit this in the near future, I apologize for that. One I just don't know how to fix, sometimes the autofill dropdown is hidden behind the keyboard. And the other, if you write `NEW_TASK` in the search box, it won't find it, you need to write `FLAGS_ACTIVITY_N`, as it is just an autofill.

I think that's all. Let me know what you think!


PabloOQ:
I forgot to mention, I added the module just before the debug module because I felt this module is kind of a debugging module.

TrianguloY:
Lots to unpack! But it's incredible the work you made, well done! 

From a quick review I see a few things I think can be simplified (mostly java-related), a couple I'll probably change (that global state goes in the right direction, but I think it can be implemented easier and less static-like) and a lot that I really like (like using reflection to make the available flags).

It'll take probably a couple weeks to review and clean the code, but be sure that I'll check and explain all decisions. In the meantime, you think you can fix the failed build? You used a newer java feature not available in older supported androids. I didn't check how it's used but it can probably be changed. If you are not sure how don't worry as I'll check it while reviewing in any case.

Also don't worry for those fixme, it's perfectly ok for me so I can quickly see the issues and try to fix them myself, if possible (I personally think it's better than don't say anything).

Thanks again!

TrianguloY:
Ok, status update!

I've reviewed all the code except the module itself (all classes except FlagsModule and related xml) and made a few changes to simplify or simply regroup existing features with new utilities. The main change is that the global data is now a standard map without any extra interface, all managed directly by AModuleDialog. I think it looks a bit cleaner, suggestions are welcomed.

I'll review the module itself another time, so far looks promising!

PabloOQ:
> The main change is that the global data is now a standard map without any extra interface, all managed directly by AModuleDialog. I think it looks a bit cleaner, suggestions are welcomed.

Indeed, looking at it now that seems like the obvious (and straight forward) way of implementing it.

TrianguloY:
Finally! I've reviewed the flag module, here is my thoughts:

The module is pretty well done and configurable, I've refactor the code a bit because that's what I like, but there were no big changes other than a few simplifications. My main concern is that it's a bit hard to use for a normal use, take for example the custom tabs button: it's just a 'press once' toggle. Here you need to write the flag name, then press add, or expand and then press remove. But, since you wrote it as a 'debug' module, I think that qualifies for its inclusion.
To be honest I have a couple ideas to create an easier-to-use flags module, the main difference is that mine is still only an idea, yours is a fully operational implementation. For this reason I'll merge it now, and maybe in the future I'll replace/add the improvement. I can also write here how I plan to, in case you want to try yourself (if so tell me and I'll explain).

The other main reason is that I have a couple improvements on my local branch that change a lot of the internal module code, and that I can't add yet until yours is merged (otherwise the number of conflicts would be noticeable).

So: merged!

PabloOQ:
> My main concern is that it's a bit hard to use for a normal use, take for example the custom tabs button: it's just a 'press once' toggle. Here you need to write the flag name, then press add, or expand and then press remove.

Mmmm, I understand, but there are so many flags, I don't know if there is a less clunky way of doing it. Best I could come up with is a list of all flags and a checkbox on the side, but that takes up way more space.

>  I can also write here how I plan to, in case you want to try yourself (if so tell me and I'll explain).

I would love to! But I don't think I'll be able to implement it, not soon at least, as I will be busy for the upcoming weeks.

> The other main reason is that I have a couple improvements on my local branch that change a lot of the internal module code, and that I can't add yet until yours is merged (otherwise the number of conflicts would be noticeable).

If you think I can be of any help, you could upload it as a branch, do a PR and ping me in there, with some modules as example I can try to adapt others to the new format. Although, as I mentioned earlier, I don't think I'll be able to find time in the next few weeks.

PR #217, opened 2023-04-12
Fully translated into Turkish

symbuzzer:
The English strings.xml file has been re-translated. Previous translations checked and necessary corrections made.


symbuzzer:
?

TrianguloY:
There are some issues with the translation, for example you can't use ' without escaping it as \' 
that's why the build fail.

But don't worry, I'll probably be able to fix them myself before merging, you can leave it as is.

symbuzzer (commented):
There doesn't appear to be any problems for now. After seeing the strings on the application, I plan to revise them again. Thanks...

Issue #98, opened 2022-09-16
Add copy url button


User1l0:

TrianguloY:
You can long press the share button, which does copy the url to the clipboard (I think that's what you asked for, isn't it?)

I remember this was discussed somewhere, to separate the open and share functions which will make more space to add extra buttons. I don't see an issue with it so I'll keep this.

Ilithy:
> I remember this was discussed somewhere, to separate the open and share functions which will make more space to add extra buttons.

Yup, on #37 and #47

TrianguloY:
> issue can be closed

Hmm, the issue is not yet implemented. Why do you suggest to close it?

TrianguloY:
Correct, I explained it on my comment.
However I decided to keep the issue open because the idea is to add a dedicated separate button. Adding functionality to a button long press allows for compact UIs, but it's hard for discoverability.

TrianguloY:
I've never said I was going to remove that functionality, but rather add a separate button. I'm also glad to hear you like it, I added it for myself and without any way to track the users usage it's hard to know if a functionality is used or not.

In any case it's still something to think about, that won't probably change in the near future.

PR #233, opened 2023-05-15
I have added issue #98, which was the request to add a 'copy URL' but…

bjbjorn:
…ton in the modules screen.

I have also added a button to the home screen that checks the URL in the clipboard. This button used to be hidden behind the three dots, which seemed like a really weird location to me. Considering this is always the way I use the app.


TrianguloY:
I think the last merge wasn't made correctly, the build fails because there is some old code left.
I can try to fix it once I review this, but if you can fix it it would help a lot!

bjbjorn:
I think it's fixed now.
FYI I'm doing this as part of a school assignment, which is to contribute to an open source software project to learn working with GIT. I'm sorry if I'm causing you any trouble or creating unnessecary work for you.

TrianguloY:
Oh! That's a nice assignment.

As for me, don't worry at all. It's no trouble, in fact I should be the one apologizing for not being able to review this sooner. Today or at most tomorrow you'll have a proper review, promise!

TrianguloY:
Reviewed! Here are all my comments.

Code wise it's a nice implementation. It wasn't very hard, but usually short and direct prs are the best both for the developer and the reviewer.
You even kept the code style and made the correct required changes. The only 'mistake' was that you didn't specify an id for the new MERGEBUTTON_PREF.
The rest is ok, although I refactored it a bit (just because I prefer it that way, I do this with almost all prs so it's nothing special). 
Also, when using git you can check old code at any time, so instead of commenting unused code it's usually better to just delete it.

The git usage was good, although there are a couple things that can be improved. These are all my own personal opinions, others may disagree:
- The PR contained two different things: adding the copy button to the dialog and adding the shortcut button to the main screen. It's usually better to use two different PRs, but if you prefer a single one at least try to keep a commit for each of them (instead of all in a single one) which can help later when checking or editing them. The extra commit with the toggleable suggestion as a separate commit was the right approach. And for the fixed merge you should have changed the commit comment (or make and amend, but that is advanced git and difficult if you are still not used to it).
- When writing a git comment you can write anything, but the convention is to use the first line as a 'title' (80 chars max aprox) then a double line break and below the full description. Most tools make this distinction when showing the commits so it's a good idea to follow it. Your commits are a single block of text, so for example here in github they are cut. There are also other conventions to use 'prefixes' or even emogis to categorize them, but I personally don't follow that.
- [personal opinion that others may disagree]: When you make a PR that uses an old commit as a parent, it's better to keep it that way until the reviewer asks for a merge or a rebase. If a merge is required because there are conflict I personally prefer to make a rebase instead, if there are no conflicts it's better to leave it.
- For GitHub only: if your commit fixes/closes a specific issue you can link it so that [it automatically closes when merged](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests). It can be closed manually later, so not a big deal, but nice to know.
- When doing a PR with changes involving an UI, it's a good idea to add screenshots to the PR description.

Feature-wise, here are my thoughts:
- I added the shortcut to the main screen just because it felt nice, but the shortcut is available as a static shortcut that you can use from you home screen, so if your usercase is to click it, you can use it instead of opening the app (or even use the quick settings tile). Having said that, your implementation was correct (although you could have used a borderless button) but you are right that requiring two clicks was unnecesary. I restored the old button but made it so that it is always displayed as an icon. I think it looks better, but you can discuss it if you don't agree.

image

- The new copy button is ok, you even followed the same style. The problem is that this new button is taking too much space, and now the open button looks bad on small screens:

image

I think this could be a perfect oportunity to replace the share button (and the new copy too) with a share/copy icon. I can easily add it, but I'll leave it to you (unless you can't, just say so and I'll add it myself, no problem).

This is roughly what I'm suggesting: 

image

For the icons, the easiest is to create them with Android Studio own icon creator (File->New->Vector asset), 'content copy' and 'share'. You can check the custom tabs and incognito buttons/icons for reference. Don't forget to call AndroidUtils.longTapForDescription for the contentDescription as long tap! (except for the share button in merged mode of course)

Well, I think this is all. Try to make the suggested icons change (or tell me to do it), afterwards I'll merge it.
And if you have any question just ask!

TrianguloY:
Hi @bjbjorn, will you be able to work on this this week?
I have other code to upload, but I was waiting for this pr to avoid more merging. I can either upload the other code (but then you'll need to merge again) or implement this features myself and close the pr (which I don't mind and probably is the better one).

As you prefer. I'll go for the second option (implement and close this pr) on Saturday if you haven't reply until then.

PabloOQ:
While we are at it with consistency changes, I think adding a preference to close the dialog when copying would be a good addition.
#42 #109

bjbjorn:
Hi, sorry for the late answer i've been busy (and still am). Your idea looks a lot better, and I should have thought of that myself tbh. You may add it yourself if you haven't already. As my finals are coming soon and have some other assignments to work on. Also thanks for the feedback it was really useful.

TrianguloY:
No problem! I'll continue from here then.
Thanks for the original implementation, and good luck on your finals!

PR #243, opened 2023-06-04
Pattern Checker: Fix optional groups giving a null instead of an empty string

### LINKED ISSUE #237: Pattern Checker: Optional groups yield "null" instead of an empty string
### Steps to reproduce

1. Add the following into the Pattern Checker config:
```json
[
  "Optional groups": {
    "regex": "^https:\/\/test\\.invalid\/(.+)?$",
    "excludeRegex": "^https:\/\/test\\.invalid\/exclude\/",
    "replacement": "https:\/\/test.invalid\/exclude\/$1"
  }
]
```
2. Open https://test.invalid/
3. Apply the pattern

### Expected behavior

The URL gets transformed into https://test.invalid/exclude/

### Actual behavior

The URL gets transformed into https://test.invalid/exclude/null

### URLCheck version

2.11

### Android version

Android 9

### Android Custom/Specific ROM or Device

OxygenOS 9.0.10

### Other details

I encountered this bug through this monstrosity that adds `?useskin=vector` to links for wiki.archlinux.org:
```json
[
  "Use old skin for Arch Linux Wiki": {
    "regex": "^https?:\/\/wiki\\.archlinux\\.org\/([^#?]*?)(?:\\?([^#]+))?(#.+)?$",
    "excludeRegex": "^https?:\/\/wiki.archlinux.org\/.*[?&]useskin=",
    "replacement": "https:\/\/wiki\\.archlinux\\.org\/$1?useskin=vector&$2$3",
    "automatic": true
  }
]
```

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

the-blank-x:
Fixes #237

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
@@ -74,6 +74,7 @@ class PatternConfig extends AModuleConfig {
                         R.string.mPttrn_userContent,
                         "https://github.com/TrianguloY/UrlChecker/wiki/Custom-patterns"
                 ));
+        RegexFix.attachSetting(views.findViewById(R.id.regex_fix));
     }
 
 }
@@ -85,12 +86,14 @@ class PatternDialog extends AModuleDialog {
     private LinearLayout box;
 
     private final PatternCatalog catalog;
+    private final RegexFix regexFix;
 
     private final List<Message> messages = new ArrayList<>();
 
     public PatternDialog(MainDialog dialog) {
         super(dialog);
         catalog = new PatternCatalog(dialog);
+        regexFix = new RegexFix(dialog);
     }
 
     @Override
@@ -159,7 +162,7 @@ class PatternDialog extends AModuleDialog {
 
                     if (replacement != null) {
                         // replace url
-                        message.newUrl = RegexFix.replaceAll(url, regex, replacement);
+                        message.newUrl = regexFix.replaceAll(url, regex, replacement);
 
                         // automatic? apply
                         if (data.optBoolean("automatic")) {
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
@@ -1,5 +1,7 @@
 package com.trianguloy.urlchecker.modules.list;
 
+import android.annotation.TargetApi;
+import android.os.Build;
 import android.view.View;
 import android.widget.Button;
 import android.widget.LinearLayout;
@@ -24,6 +26,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Random;
 import java.util.regex.Pattern;
+import java.util.regex.Matcher;
 
 /**
  * This module checks for patterns characters in the url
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
```diff
@@ -225,4 +228,100 @@ public Message(String pattern) {
...
+    private static int appendReplacement(String text, Matcher matcher, StringBuffer sb, String replacement, int appendPos) {
+        // Copied from https://android.googlesource.com/platform/libcore/+/refs/heads/android13-release/ojluni/src/main/java/java/util/regex/Matcher.java#714
+        sb.append(text.substring(appendPos, matcher.start()));
+        appendEvaluated(matcher, sb, replacement);
+        int newAppendPos = matcher.end();
+
+        return newAppendPos;
+    }
+
+    @TargetApi(Build.VERSION_CODES.O)
```
TrianguloY, line 263: Is this necessary? 🤔

app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
```diff
@@ -225,4 +228,100 @@ public Message(String pattern) {
             this.pattern = pattern;
         }
     }
+
+    /**
+     * On Android 10 and under, optional groups may yield a "null" in the replacement output instead of an empty string.
+     * Therefore, we just copy the implementation from newer versions
+     * https://github.com/TrianguloY/UrlChecker/issues/237
+     */
+    private static String replaceAll(String text, Pattern pattern, String replacement) {
```
TrianguloY, line 237: Is it possible to place an early-return-if here to call the original unmodified replaceAll if we are on an android version where the fix is already native?

app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
```diff
@@ -158,7 +161,7 @@ public void onModifyUrl(UrlData urlData, JavaUtils.Function<UrlData, Boolean> se
 
                     if (replacement != null) {
                         // replace url
-                        message.newUrl = url.replaceAll(regex, replacement);
+                        message.newUrl = replaceAll(url, regex, replacement);
```
TrianguloY, line 164: Since this is an (impressive) hax, please move all the new logic to a new class (maybe utils.RegexpFix? Not sure).
In any case, this line here should call that class, and that's it.

TrianguloY (changes requested):

app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
the-blank-x: `lintAlpha` gave me an error for not doing so

app/src/main/java/com/trianguloy/urlchecker/modules/list/PatternModule.java
the-blank-x: Sure

TrianguloY:
Perfect! Thanks! 
Just let me check a few things first with my computer (and also I want to see how you fixed the error I had). I'll tell you if I need anything else.

Good job!

TrianguloY:
Finally!

I was able to review this. Apparently my version failed because I wasn't copying all the required code, like yours.

I validated it, but since this is still a big hack, I added a setting to disable it (just in case, the fix is enabled by default).

Again, thank you! Nice work! 

merged

Issue #224, opened 2023-05-08
Option to hide modules while keeping them enabled


PabloOQ:
### Describe a related problem (optional)

I usually do not use the "Status code" module but I like to keep it enabled for the times I need it. I feel the same with some other modules too, so I might have 3 modules that I rarely use, but that are very useful when I need them. When those are not needed the URLCheck dialog is unnecessarily big and cluttered.

### Describe your suggested feature

A way to hide these modules in an overflow menu like the one in the "Queries remover" or "Flags editor". Maybe even have the visibility be decided by a regex matching the URL.

### Describe alternatives you've considered for your suggested feature

Disabling the modules.
Moving these under the "Open with" module, so I know that everything that is always relevant is in the upper half of the dialog, and circumstantial modules are in the lower half.

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

TrianguloY:
In the newer update (as you probably saw) modules will be hidden when you can't interact with them and they don't have any information to show.

Collapsing them to a dropdown each is probably counterproductive, since all modules that 'are just there' are very small (if there is one that isn't we can try to reduce that one).

Maybe a global dropdown to collapse all there?

Or wait, what about a 'separator' so that all modules under it will be hidden by default in the separator toggle? This is probably very easy to implement, it will be just a normal module with a special feature for the window, and it will also be easy to setup (just move them below, as you describe!)

PabloOQ:
> Maybe a global dropdown to collapse all there?

Certainly! That was what I meant.

> Or wait, what about a 'separator' so that all modules under it will be hidden by default in the separator toggle? This is probably very easy to implement, it will be just a normal module with a special feature for the window, and it will also be easy to setup (just move them below, as you describe!)

I was thinking about this too, but it didn't feel right to me. However, you are right, that is probably the easiest option and seeing that we can't implement much of the proposed features because lack of time, maybe that's the best way.

TrianguloY:
```java
t=new Intent(Intent.ACTION_VIEW,Uri.parse("http://google.es"))
t.putExtra("private_browsing_mode",true)

cntx.startActivity(t)```

It works! I'm using the latest play store Firefox version, and I can confirm that the code above works! (even if UrlChecker is opened and then you choose Firefox, thanks past me for keeping the original intent)

[...]

PR #252, opened 2023-06-24
New Drawer module

PabloOQ:
Solves #224.

Not much to explain really, modules that can be hidden are in a separate layout that can change its visibility. The module itself has a button to toggle the visibility.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/dialogs/MainDialog.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/dialogs/MainDialog.java
@@ -28,6 +28,7 @@ import com.trianguloy.urlchecker.utilities.Inflater;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -47,7 +48,7 @@ public class MainDialog extends Activity {
     /**
      * All active modules
      */
-    private final Map<AModuleDialog, List<View>> modules = new HashMap<>();
+    private final Map<AModuleDialog, List<View>> modules = new LinkedHashMap<>();
 
     /**
      * Global data to keep even if the url changes
@@ -125,7 +126,10 @@ public class MainDialog extends Activity {
             }
 
             // third notify for final changes
-            for (var module : modules.keySet()) {
+            var mods = new ArrayList<>(modules.keySet());
+            // drawer module needs to display AFTER the hidden modules
+            Collections.reverse(mods);
+            for (var module : mods) {
                 // skip own if required
                 if (!urlData.triggerOwn && module == urlData.trigger) continue;
                 try {
@@ -345,6 +349,16 @@ public class MainDialog extends Activity {
 
     public void toggleDrawer() {
         setDrawerVisibility(getDrawerVisibility() == View.GONE ? View.VISIBLE : View.GONE);
+    }
+
+    public boolean anyDrawerChildVisible(){
+        int childCount = ll_hidden_mods.getChildCount();
+        for (int i = 0; i < childCount; i++) {
+            if (ll_hidden_mods.getChildAt(i).getVisibility() == View.VISIBLE){
+                return true;
+            }
+        }
+        return false;
     }
 
     /* ------------------- its a secret! ------------------- */
--- a/app/src/main/java/com/trianguloy/urlchecker/dialogs/MainDialog.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/dialogs/MainDialog.java
@@ -28,7 +28,6 @@ import com.trianguloy.urlchecker.utilities.Inflater;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -48,7 +47,7 @@ public class MainDialog extends Activity {
     /**
      * All active modules
      */
-    private final Map<AModuleDialog, List<View>> modules = new LinkedHashMap<>();
+    private final Map<AModuleDialog, List<View>> modules = new HashMap<>();
 
     /**
      * Global data to keep even if the url changes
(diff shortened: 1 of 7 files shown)


TrianguloY:
Looks good! Although I wonder if it could be possible with just one linear layout, maybe it was harder to manage the modules own visibility? Probably

I played with it a bit and it behaves as expected, but I think it would need to detect when there are no visible modules on the drawer, to hide it (otherwise clicking the arrow does nothing).

Another improvement would be to replace the arrow with something else (maybe two arrows?) It doesn't look bad, but it feels like it's missing something...

I'll test some things, but feel free to try yourself too.

Thanks!

PabloOQ:
> Didn't see it, but I guess two arrows would look weird. Maybe Expand Drawer and Collapse Drawer instead? Or Open/Close, or Show/Hide, or whatever, but with text?

I like it, but from other PRs it seems like we are moving away from text in favor of icons.

> Another improvement would be to replace the arrow with something else (maybe two arrows?) It doesn't look bad, but it feels like it's missing something...

What about this?



Or maybe using three horizontal dots (meatball, horizontal ellipsis)?

> Looks good! Although I wonder if it could be possible with just one linear layout, maybe it was harder to manage the modules own visibility? Probably

Yeah, I tried that. Because the modules are responsible of its own visibility they needed to know if they were in the drawer and the state of the drawer, it would also require tweaking each one to add extra logic in their respective onDisplayUrl(), aside from that each one would need to be notified when opening or closing the drawer.

> I played with it a bit and it behaves as expected, but I think it would need to detect when there are no visible modules on the drawer, to hide it (otherwise clicking the arrow does nothing).

I have a fix on my local branch now!

TrianguloY:
I prefer to use icons, yes. It's more of a personal decision, but I think it also has benefits related to translations (an icon is more 'universal', and even though the icon must have a text description, it's less important). I'm open to discussion though.

Anyway, your second screenshot seems more cluttered. What about these? Ranked from top (my favorite) to bottom (less favorite)

imageimage

imageimage

imageimage

TrianguloY:
On the three options the full row is clickable, the icons placement is just visual.

Even in that case, the first one does looks more "clean", I agree.

PabloOQ:
I pushed the fix, no changes in the UI there.

> Ohh, I like the first one! Both hands can easily reach the buttons, and it's still clean.

> But I guess it can be a bit hard to press

This is why I wanted to put the border of the button, so the user clearly knows the whole module is a button and can comfortably press the button.

> Ranked from top (my favorite) to bottom (less favorite)

I also like the first one.


Later I´m going to try something that I think it could work, if not, the best option is probably triangulo's first schreenshot.

TrianguloY:
I...don't like the full button border, I prefer it to be invisible. Or maybe tint all the background with a 25% or so transparency? I need to check that.

What I really like is the horizontal line, in fact I tried to do it but didn't know how to do it properly, how did you make it??

For the muted colors, I have tried to make everything that is interactable colored, and everything that isn't grayscale (not everything follows this logic, I know, but most of it does and my idea is to change what doesn't). That button feels like it's disabled somehow, maybe it's just me.

PabloOQ:
> What I really like is the horizontal line, in fact I tried to do it but didn't know how to do it properly, how did you make it??

Well, I wouldn't say I did it properly, I was hoping to get it sorted out later if we went that way😅. The height of the separator is forced to 3px, and to keep it centered is just a bunch of space.
image


> That button feels like it's disabled somehow, maybe it's just me.

Agh, you are right, didn't think of it but it does give that feel indeed.

TrianguloY:
Oh! That seems much easier, I though the line was inside the button somehow. In any case that may be even easier, I don't think you need any extra space views, just set the gravity to center and apply some padding.

Btw, I won't be able to work on the app for a couple weeks, and I would like to review properly before merging, sorry. I just published version 2.12 to avoid taking more time, but on my return this is the first thing I'll check. If you want to develop other changes you can either create a new branch from master (in parallel, so the drawer code won't be there) or just create it from this one (in series, as if this one was master). In any case I can later merge/rebase as necessary, so don't worry about that and just do whichever you are more comfortable with. Great job in any case! and sorry again

TrianguloY:
And the last tests I wanted to make. I prefer the first, but I don't dislike the second one.

imageimage

imageimage

PabloOQ:
> Btw, I won't be able to work on the app for a couple weeks, and I would like to review properly before merging, sorry.

No worries! I'll move on to the incognito issue then.

> And the last tests I wanted to make. I prefer the first, but I don't dislike the second one.

I prefer the second one. The first one... I just can't unsee it, to me it looks like an emoji `>_<`

TrianguloY:
Ok, so... Hi! I'm back.

Let me be honest, I don't like how the drawer module requires 'special' features, maybe it should have been better to add it as a built-in feature, I don't know. I still like it being a module though.

I tried to extract the logic, move it to a different file, even using callbacks; but everything was just worse because your solution is both simple and easy.

So, I just made some minor renames and cleanups, and made a special callback for the visibility problem (I don't like the modules to be ordered, they were in the past and it had a lot of issues) and kept your logic.

I'll test this version on my device for a couple days, if I don't see anything strange I'll merge it!
Anyone is encouraged to test it and give their opinion too.

PabloOQ:
> Let me be honest, I don't like how the drawer module requires 'special' features, maybe it should have been better to add it as a built-in feature, I don't know. I still like it being a module though.

I feel the same but it required minimal changes. At the very least it needs a way to be visualized in the modules screen, that would mean either add it as a module or make changes to the module screen, which, among other things, I felt it would take more time.

> So, I just made some minor renames and cleanups, and made a special callback for the visibility problem (I don't like the modules to be ordered, they were in the past and it had a lot of issues) and kept your logic.

I did two versions of this, as I wasn't conviced either on sorting the modules, but the other one, while minimal, required changes in all modules, as I wanted to reuse the module ids, but that is currently not possible from the `AModuleDialog` class. You can check it out [here](https://github.com/PabloOQ/UrlChecker/tree/separator-mod-id) if you want (there is a stream that needed changing as it should not work on lower android versions).

Btw, I'm still working on the incognito issue, but don't expect it soon as it is taking me some time.

TrianguloY:
After some daily usage, it is stable and working.
No one (if any has tested) hasn't said anything either, so...

Merged! 🥳

Issue #175, opened 2023-02-06
Export/Import app configuration


Ilithy:
**Describe you'd like**
it would be very practical, with the extent of the functionalities, and the quantity of parameters present in URLCheck... to have a function of saving the configuration of the application.

**Additional context**

- It would be nice if the backup could include the log (generated by the logging module).

- It would be perfect if the inclusion of the log in the backup file could be configurable. (for those who do not wish to keep it, those concerned about the size of the backup file, or other reasons)

- it would be great if the backup include advanced editors configuration (for URL Cleaner, Pattern Checker and Hosts labeler module)

Thanks a lot.

Ilithy:
Thank you very much, please take the time

TrianguloY:
This is finally implemented (it only took 1 year!).

I know it took a lot of time, and maybe is not as powerful as expected, but for a first implementation I'm quite happy with it. The update is now live on the playstore as beta, will promote to stable in the following week if nothing bad happens.

I may improve, and add more settings, if people requests it. But for now I'll wait for bug reports (hopefully none!) and tweak as necessary.

I'll close this issue as completed, but feel free to create new ones for improvements or (specially) bugs.

PR #311, opened 2024-01-28
Backup/restore (WIP)

TrianguloY:
Even though it's still a work in progress, and things may (and probably will) change, I'm creating this PR for people to try it if you want (check the automatic bot reply).

Any comment will be greatly appreciated!

This will close https://github.com/TrianguloY/UrlChecker/issues/175


PabloOQ:
First I should mention that I haven't looked at the code so I can give proper feedback as a user and not a dev.

I have tested all 3 versions and it didn't work on my phone. When doing the backup it shows a toast "Can't create backup", however it doesn't explain why, leaving me totally clueless on the reason on why it can't. I looked in the android settings of the app and it has the same permissions as the release version (full network access), nothing about storage. In the version with the broken dialog it just stays at 0% and the dialog can't be closed.

The restore button does show me a file explorer to choose files, because I can't create a backup I could only test with the wrong file, I tested a random image from my gallery, it correctly shows a toast "Unable to restore the backup file", although I don't know if this is because it is not a .backup or because it has no permissions to read the file.

Deleting through the advanced options seemed to work correctly, I tested reordering modules, disabling them, some options inside the modules, modify some files and the tutorial pops again too.

TrianguloY:
> First I should mention that I haven't looked at the code so I can give proper feedback as a user and not a dev.

That's exactly what I need right now, thanks! 


> I have tested all 3 versions

> In the version with the broken dialog it just stays at 0% and the dialog can't be closed.

Just need to check the last version, the others have issues or just lack improvements.


> and it didn't work on my phone. 

:O that's unexpected...but better to detect now than later! Can you remind me which android version do you have?

> When doing the backup it shows a toast "Can't create backup", however it doesn't explain why, leaving me totally clueless on the reason on why it can't. 

Exception management is not yet implemented, although it's difficult to do in this case. I'll try to improve it though.
For now you can try to check the logcat, there you should find the good-old red stacktrace with more info.

> I looked in the android settings of the app and it has the same permissions as the release version (full network access), nothing about storage.

According to android (and tested on a couple devices) you don't need any permission to create a file in the downloads folder. Which is strange...I know


> The restore button does show me a file explorer to choose files, because I can't create a backup I could only test with the wrong file, I tested a random image from my gallery, it correctly shows a toast "Unable to restore the backup file", although I don't know if this is because it is not a .backup or because it has no permissions to read the file.

It's expecting a zip file, so that's probably the reason. You can test with any zip (it should just not restore anything) but I'll try to upload a sample zip here so you can test. I'll also try to improve the error messages here too.

> Deleting through the advanced options seemed to work correctly, I tested reordering modules, disabling them, some options inside the modules, modify some files and the tutorial pops again too.

Well, at least something is working. But the delete option does not imply zips at all so...that's probably the reason.

TrianguloY:
OK, so, you do need the permission for some specific android version (between 6.0 and 11).
I'll add the popup to grant it next, but for now if you manually add it from the settings it should hopefully work!

PabloOQ:
> According to android (and tested on a couple devices) you don't need any permission to create a file in the downloads folder. Which is strange...I know

Not so weird, I think Firefox Desktop does the same with addons, but I might be wrong, I know I have seen this behaviour somewhere, in fact I was suspecting this was the case here too, still I wanted to check.

> Can you remind me which android version do you have?

Yeah, about that... I have been stuck with a Huawei phone (kind of ironic considering I joined this project mainly because of privacy with the ClearURLs module) since I broke my last phone, and this one is from the "bad batch", when Huawei was banned from using google services, which translates to "a lot of broken stuff because missing libraries", maybe it's because of that? Although, looking at the error it doesn't seem that way. In any case, android 10, EMUI 10.1.0. 

> It's expecting a zip file, so that's probably the reason. You can test with any zip (it should just not restore anything) but I'll try to upload a sample zip here so you can test. I'll also try to improve the error messages here too.

Check this out, coincidentally the GitHub download link for the testing APK is... a .zip file! So of course I had to test it with the app itself, I honestly find it very funny. Jokes aside, it says "Backup restored", and obviusly doesn't restore anything.

> but for now if you manually add it from the settings it should hopefully work!

Still it doesn't work.

```
java.io.FileNotFoundException: /storage/emulated/0/Download/UrlChecker_20240201164930.backup: open failed: EACCES (Permission denied)
    at libcore.io.IoBridge.open(IoBridge.java:496)
    at java.io.FileOutputStream.<init>(FileOutputStream.java:235)
    at java.io.FileOutputStream.<init>(FileOutputStream.java:186)
    at com.trianguloy.urlchecker.utilities.wrappers.ZipWriter.<init>(ZipWriter.java:25)
    at com.trianguloy.urlchecker.activities.BackupActivity.lambda$backup$4(BackupActivity.java:144)
    at com.trianguloy.urlchecker.activities.BackupActivity.$r8$lambda$jpT3LsUIhEHLbiRWyMFub3o39mM(Unknown Source:0)
    at com.trianguloy.urlchecker.activities.BackupActivity$$ExternalSyntheticLambda6.accept(Unknown Source:6)
    at com.trianguloy.urlchecker.utilities.wrappers.ProgressDialog.lambda$new$0(ProgressDialog.java:43)
    at com.trianguloy.urlchecker.utilities.wrappers.ProgressDialog.$r8$lambda$mHOfL2_pyMAAGoscqxjichzblcc(Unknown Source:0)
    at com.trianguloy.urlchecker.utilities.wrappers.ProgressDialog$$ExternalSyntheticLambda0.run(Unknown Source:4)
    at java.lang.Thread.run(Thread.java:929)
```

TrianguloY:
> Yeah, about that... I have been stuck with a Huawei phone (kind of ironic considering I joined this project mainly because of privacy with the ClearURLs module) since I broke my last phone, and this one is from the "bad batch", when Huawei was banned from using google services, which translates to "a lot of broken stuff because missing libraries", maybe it's because of that? Although, looking at the error it doesn't seem that way. In any case, android 10, EMUI 10.1.0. 

No, I don't think so. Maybe, but it would be strange. I'll test with an android 10 emulator.

> Check this out, coincidentally the GitHub download link for the testing APK is... a .zip file! So of course I had to test it with the app itself, I honestly find it very funny. Jokes aside, it says "Backup restored", and obviusly doesn't restore anything.

Oh, right! At least we can confirm that the app can read zips, so that's something.
And yes, it doesn't restores anything because the data is not there, but it's not deleting existing data either because I didn't want to delete things just in case you are restoring a backup that was made without that specific data, even if you are selecting it on the screen. I think this makes sense, but I would like to hear other opinions.

> Still it doesn't work.

Oh :(
Let's see if the emulator can reproduce the issue.

davidj77:
Hi, I'm testing it by using a Moto phone with Android 7
I was able to create a backup by giving access to phone storage and restore it with no fails. 
Gallery_1706869138558

TrianguloY:
Good and bad news. I think I fixed the permission issue, turns out there is an official Android way to ask the user to create a file, and it can be created anywhere (it even asks to confirm if it exists!).
The bad news is that it requires api 19 (Android 4.4, Kitkat, from 2013) and current min is 14 (Android 4.0.1 – 4.0.2, Ice cream sandwich, from 2011).

I started adding the code for those older versions, but it was becoming a hassle and even failing in some situations.

So instead, I decided to increase the minSdk and remove support for those older versions. I don't like it, I feel that older devices are still capable of running most apps and if they can't it's just because developers don't spend a bit of time trying to support both. In this case however the burden of supporting the previous version is too much for me.
If there is a user using android 4.3 or less and really wants to keep using the app, please tell me and I may consider keeping the minSdk (but probably disabling the backup feature). Otherwise...that min needs to increase from time to time, and I think this is the best moment to do so.

PabloOQ:
It works now! And without requiring any extra permissions, great.

>  I decided to increase the minSdk and remove support for those older versions

I am sorry to hear that, but at least it wasn't a big jump.

TrianguloY:
With these, all the minimal required things I can think of are fulfilled, so unless a bug is discovered I'll merge this on the following days.

There are some things and improvements I would like to add, but to avoid postponing this update even more I'll release this and improve later.

Thanks for all of you that helped with testing!

PabloOQ:
I tried it yesterday and it worked as intended. I have already uninstalled it, but I have one question on the behaviour. If I want to backup ClearURLs **custom** rules, would I need to toggle the cache files?

TrianguloY:
No, it is considered user data.

ClearUrls setting has both "official" and custom rules on the same object (the idea was for the user to be able to edit the rules themselves). I could have added logic to know if the official subobject was modified or not, but to keep things simple for how I just considered it user data. This means that the backup does contains some redundant data, but as I said I wanted to keep things simple for now.

Currently, the only thing considered cache is the hosts database (the one you need to press "build" to generate, not the settings) which internally is not editable by the user. This consists on pretty big files, that's why I added a toggle for them.

Btw, if you need to check what is being saved, just open it! The backup is a zip!

PR #331, opened 2024-03-20
Update AModuleConfig.java

Suhatec:
Reverse every thing


TrianguloY:
lol

Like the idea, but you could have replaced all files in the repo (it takes literal seconds), the -1 should have been a +1, there is a typo one the last public modified...

Anyway, I don't know what are you trying with this, not sure if it is even a good idea to reply, but unless you probe this is just a harmless prank, I'm sorry but I'll need to block and report you.

In the meantime I'll just close this.

PR #338, opened 2024-03-26
 Refactor Utils class for Better Readability

nikul014:
Refactored the Utils class to enhance code readability by separating functionalities into two distinct classes:

- Created HttpUtils class for HTTP-related operations.
- Introduced LocaleUtils class for locale-related operations.
- Cleaned up code for improved readability and maintainability.


TrianguloY:
I'm currently bust with some other projects so I'm afraid it will take time (probably a couple or more weeks) until I can properly review this pr.

Although I'm really interested on it, since "Utils" are one of my favorite things so any suggestion and improvement is really welcomed :D

Expect lots of discussions! (But please feel free to discuss back. I'll provide reasons for why I do things a specific way, but I'm open to be proved wrong!)

nikul014:
Hello,

Thank you for responding back.

Can you please do it in the next two days as it is part of a graded
assignment that needs to be implemented before the 28th. So it is my humble
request to accept the pr as soon as possible.

If PR will be accepted then i get the grade in the subject else i loose
marks. I am open to discuss any changes needs to accept this PR.  Please
reply her if required.


Thank you for understanding.

TrianguloY (changes requested):
First of all: I really apologize for taking too long to review. Even though I'm lucky to have free time to spend, it is often limited and I need to divide it into several projects. This PR was a long one (shorter than I first expected, but still long) so I had to find a day where I could spend more time than usual.
As I explained originally, I really recommend to start these types of projects with a shorter scope, perhaps extracting one of the Utils class only, and create several PRs afterwards with more changes. I know this may seem to take more time, but in reality it is often faster and far far easier to review, since you can apply and adapt to the suggestions sooner.

Below you can find some of the general comments about the review. In general the pr is a mix of long class spliting (good), short methods splitting (not good) a couple rename (ok) some minor refactors (some agree, some don't) and a couple fixes (mandatory for me to add). As an overview, the pr is ok but the problem again is that you mixed several things into one. I guess it was due to it being a class assignment (and I don't know what where your requirements) so sorry if it hasn't gone as you expected. I hope this experience helped you, it did for me when I made similar mistakes in the past while being on your side.

As for the PR itself:
1. You can review my comments and reply and/or challenge them if you don't agree with something. In the end I'll have the final decision, true, but I can be wrong too (and I like a good discussion) so please try to change my mind if you really want. I will reply back until we have an agreement (I should be faster, I hope) so this will take time.
2. You can simply ignore them and let me make the changes I want. As I said there are things I agree and things I don't, but I can make the changes myself. You will still be credited as the author of the pr and the commit (I'll be a co-author) if that's what you wanted. [Important: if you prefer this, tell me so I can start as soon as possible with it].

Sorry again for taking too long to review, and thanks for you pr!

Issue #100, opened 2022-09-17
Non editable input text


TrianguloY:
The input text module is a big edittext. With all the new modules functionalities it is less necessary now to manually edit it. It is bigger than others so reducing it to just a textview could help with the clutter.
I can think of two solutions:

1) A setting to make it "editable" (enabled by default). When enabled it is just as it is now. When disabled instead of a text editor it shows the url as a normal textview (like the log module but with normal font).

2) Replace the editor with the text view as explained above, but add a "pencil" button that when clicked will go back to the current full edittext.

In not sure which one is better, will probably try both and check. (Or maybe use both with a three-option setting)

Ilithy:
>Replace the editor with the text view as explained above, but add a "pencil" button that when clicked will go back to the current full edittext.

>In not sure which one is better, will probably try both and check. (Or maybe use both with a three-option setting)


I find this idea pleasant, it avoids mishandling (touching the text does not open the editor) but still makes it easy to make changes.

and indeed an option so that the behavior of this module is identical to now, for those who prefer the old method will complete this perfectly.
A final option could be simply an unmodifiable display of the input text (really useful?)

Ilithy:
I don't know if it will be implementable as an option for this module (Imput text), but it would be really useful (I think) to have 
- An option to control the text size _(of the url)_
- An option to control the text alignment _(same)_

mariachrisochoou:
Hi! I am interested in this issue and I believe I can help. 
Can I be assigned to this issue? 
I am a first-time contributor, so any help or guidance will be very much appreciated. 

Thank you so much in advance.

TrianguloY:
> Hi! I am interested in this issue and I believe I can help. 
> Can I be assigned to this issue? 
> I am a first-time contributor, so any help or guidance will be very much appreciated. 
> 
> Thank you so much in advance.

Sure! 
The reason I haven't started this yet is that I'm not so sure if it is a good idea, or how to properly implement it. Using a dedicated button to switch from the view-only to the edit-text forms may clutter the screen, perhaps just letting the view-only text view act as a button is enough. It was mentioned that a button will avoid opening the keyboard if you press the input by mistake, but...I prefer to avoid having another extra icon.

So, I think as a first implementation just to have a non-editable textview (probably with smaller font) that when clicked will "convert" to the edit text is a good start.

I don't know what your knowledge of android development is, but feel free to ask for help and I'll try to assist whenever possible. My recommendation is:
[...]

PR #361, opened 2024-06-09
Non Editable Input Text Fix

mariachrisochoou:
This Pull Request corresponds to issue #100. 

**Goal**: Declutter Screen for Improved Usability

### Actions:
- Added a TextView as a default state for displaying the URL.
- Changed the visibility of the EditText to be initially hidden.
- Updated the TextInputModule to reflect the appropriate changes.
- Implemented an onClickListener: When the user clicks on the URL, it changes into an EditText to allow for editing.

### Testing: 
- Tested the default state of the URL display.
- Verified the visibility change of the EditText.
- Confirmed the functionality of the onClickListener to ensure the URL changes to an EditText correctly.

DefaultState

StateWhenClickedOn

### Ideas for further improvement:
- To add a switch in the input text module for the User to assign the default state (EditText or TextView).


app/src/main/java/com/trianguloy/urlchecker/modules/list/TextInputModule.java
```diff
@@ -59,8 +61,9 @@ public int getLayoutId() {
 
     @Override
     public void onInitialize(View views) {
-        edtxt_url = views.findViewById(R.id.url);
-        edtxt_url.addTextChangedListener(new DefaultTextWatcher() {
+        texttxt_url = views.findViewById(R.id.url);
+        edtxt_url = views.findViewById(R.id.urlEdit);
+        texttxt_url.addTextChangedListener(new DefaultTextWatcher() {
```
TrianguloY, line 66: This is the important bug, otherwise editing doesn't work!

```suggestion
        edtxt_url.addTextChangedListener(new DefaultTextWatcher() {
```

app/src/main/java/com/trianguloy/urlchecker/modules/list/TextInputModule.java
```diff
@@ -46,6 +47,7 @@ class TextInputDialog extends AModuleDialog {
     private final DoubleEvent doubleEdit = new DoubleEvent(1000); // if two updates happens in less than this milliseconds, they are considered as the same
     private boolean skipUpdate = false;
 
+    private TextView texttxt_url;
```
TrianguloY, line 50: can you rename the variable to just txt_url?

app/src/main/java/com/trianguloy/urlchecker/modules/list/TextInputModule.java
```diff
@@ -76,13 +79,25 @@ public void afterTextChanged(Editable s) {
                 // set
                 setUrl(newUrlData);
             }
+
+        });
+
+        texttxt_url.setOnClickListener(new View.OnClickListener() {
```
TrianguloY, line 85: You can replace this with a lambda, it's shorter

set...(v -> {
...
})

app/src/main/res/layout/dialog_text.xml
```diff
@@ -8,11 +9,20 @@
...
 
-    <EditText
+    <TextView
         android:id="@+id/url"
         style="@android:style/Widget.TextView"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
-        android:textSize="20sp"
-        android:textStyle="bold" />
+        android:textSize="18sp"
```
TrianguloY, line 17: No need for this, the appearance should be enough (same below)

TrianguloY (commented):
Short and to the point, perfect for a first time contribution!

I've played a bit with it (after fixing the listener bug) and I'm still not sure if it's worth it to have two different views for the editable/nonEditable states, instead of a single one with different styles. But for now two are ok, let's keep them.

There is however one important thing to change, right now both styles are almost identical, which makes the distinction a bit useless. The original idea was to have the non-editable view smaller, and the editable bigger. It would be nice to have it configurable, but for now you can just set the EditText style to DeviceDefault.Large, that should be enough.

Another thing that I think should be done is to go back to the non-editable view if the url changes. This is already handled by the app, you just need to add the setVisibility calls in onDisplayUrl. It would also be good to hide the keyboard, but that's a bit more difficult so I can add it later myself.

And finally, it would be a good idea to add the option to choose if it should be switcheable or always editable. This may be more complex if you are not familiar with the app code, so don't worry if you prefer me to add it later.

And I think that's it. Better to start with something simple and modify later. Thank you for the pr!

As for the modifications, as I always say: you can either do them yourself (or explain why you think it's better not to do something the way I said) or let me do them myself, no problem at all. I'll await for your decision!

app/src/main/java/com/trianguloy/urlchecker/modules/list/TextInputModule.java
mariachrisochoou: Thank you, I will apply it as soon as I can!

mariachrisochoou:
Thank you for your prompt and detailed review! I sincerely appreciate the time you've taken to provide feedback.

I've made the changes you suggested. I hope these adjustments address your concerns effectively.

Regarding the remaining issues, I regret to inform you that due to ongoing exam commitments, I won't be able to address them immediately. I aim to be able to complete these changes by the end of June. If this timeline doesn't align with your expectations, please let me know, and I'll do my best to accommodate.

Once again, thank you for your valuable feedback.

TrianguloY:
That's enough, thanks! 
I'm merge it tomorrow if possible, and don't worry about the other changes. I've been testing it on my device and I want to tweak it a bit, so I'll just continue myself. This is perfect for a first simple iteration.
Thanks again!

PR #393, opened 2024-11-03
Added webhook module for URL forwarding

anoop-b:
Added new module to send cleaned URLs to webhook endpoints.

Sample Payload:
```JSON
{
      "url":"http://www.google.com/?ref=referrer&foo=bar#tag"
      "timestamp":"2024-11-04T00:06:28Z"
}
```


TrianguloY:
Impressive! 

I've glanced over the code and it looks pretty good. I do have some minor suggestions and simplifications, but I'll post them after a proper review (hopefully this weekend...), I can also change them myself if needed, but we'll discuss this then.

As for the functionality itself, at first I was a bit confused of who would want something this specific (since you need a server that accepts this format), but then I realized that with a bit more configuration it could be possible to send it to existing webhooks, like a discord server, or a slack channel! This could be a very useful addition as some sort of external logging.


Thank you for the pr! As I said I'll try to review it this weekend, but feel free to ping me if I haven't said anything next week.

anoop-b:
Thank you. 

As for the motivation, I see this as the fastest/simplest way to get the URLs out of my device, without any additional user interaction.

For now, I have a cloudflare worker deployed that just stores the URLs in D1. 

For me this potentially replaces any browsing history/link sharing apps that I use. As I control the server side, I can then send it to any destination that I prefer. It's also the reason why I have kept the payload as simple as possible.

One good addition that I can think of right now is to add support for webhook signatures.

app/src/main/java/com/trianguloy/urlchecker/modules/list/WebhookModule.java
```diff
@@ -0,0 +1,191 @@
...
+        showStatus(R.string.mWebhook_sending, false);
+        
+        executor.execute(() -> {
+            try {
+                JSONObject json = new JSONObject();
+                json.put("url", currentUrl.url);
+                json.put("timestamp", new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'", Locale.US)
+                    .format(new Date()));
+                
+                URL url = new URL(webhook);
```
TrianguloY, line 134: I have a method in HttpUtils that should do this: performPOST
...but it doesn't work.
I guess it's because that method is not setting the content-type nor the requestMethod (apparently virustotal doesn't really care...)

In the virusTotal v3 branch I do have a Connection wrapper that will probably solve this. For now you can leave this code, but could you move it to the HttpUtils class? a `public static boolean performPOSTJson(String url, String body) throws IOException {...}` will suffice

app/src/main/java/com/trianguloy/urlchecker/modules/list/WebhookModule.java
```diff
@@ -0,0 +1,191 @@
...
+            }
+        });
+    }
+
+    private void showStatus(int stringId, boolean isError) {
+        getActivity().runOnUiThread(() -> {
+            statusText.setText(stringId);
+            statusText.setVisibility(View.VISIBLE);
+            
+            if (!isError) {
```
TrianguloY, line 163: Is this really needed? I would prefer to avoid showing an extra toast, if the module already notifies the failure

app/src/main/java/com/trianguloy/urlchecker/modules/list/WebhookModule.java
```diff
@@ -0,0 +1,191 @@
...
+
+            } catch (Exception e) {
+                showStatus(R.string.mWebhook_error, true);
+            }
+        });
+    }
+
+    private void showStatus(int stringId, boolean isError) {
+        getActivity().runOnUiThread(() -> {
+            statusText.setText(stringId);
```
TrianguloY, line 160: You can use AndroidUtils.setRoundedColor with R.color.warning (or even R.color.error) if there was an error. (AndroidUtils.clearRoundedColor to clear it)

app/src/main/java/com/trianguloy/urlchecker/modules/list/WebhookModule.java
```diff
@@ -0,0 +1,191 @@
...
+
+    @Override
+    public void onInitialize(View views) {
+        statusText = views.findViewById(R.id.status_text);
+        views.findViewById(R.id.send_button).setOnClickListener(v -> sendToWebhook());
+    }
+
+    @Override
+    public void onDisplayUrl(UrlData urlData) {
+        currentUrl = urlData;
```
TrianguloY, line 110: I think you forgot to call `statusText.setVisibility(View.GONE);` here?

TrianguloY (changes requested):
OK! review is done.

I left some comments for things I think that should be changed. Feel free to review them and answer.
As for the changes I request, you can either make them yourself, reply why you think they are not needed (if you don't agree with me, which is fine don't worry) or just agree but you prefer me to do the changes myself (which I'll happily do).

You can also reply to this review with a single post if you don't mind me doing all the changes and then merging (probably the easiest for you...but you will miss the discussion :)

Again, thanks for the pr! It's something that I'll probably have never imagined to add, but I'm sure it will be useful (it will for me!)

anoop-b:
> You can also reply to this review with a single post if you don't mind me doing all the changes

Works for me. I agree with your suggestions.

TrianguloY:
OK, it should be ready now. I'll merge it.
Thanks again!

Just one final question: how do you want to be credited for this? (if you want to be credited, that is)
Is `anoop-b` ok?

anoop-b:
Hello, thanks for merging. Any plans for release?

TrianguloY:
Sorry for the lack of news 😞. I had a lot of plans for the Christmas holidays, but life had others for me (nothing serious, don't worry).

I wanted to publish the next version with the new json editor screen, and I also still need to review Pablo's pr (which I'm still not sure how to approach...). But it's true that right now my time is more limited than never. Maybe I'll just publish a new version with whatever there is available right now (including your new module) and leave the json feature for the future. Probably better for all.

Let me put it as an own-deadline: a new version will be released before the 19th. Hopefully this will give me enough motivation to finally do it!

Sorry again for the delay, but also thanks for asking! Perhaps that's what I needed...

PR #395, opened 2024-11-10
Imold patch 1

Imold:


TrianguloY:
Hmm, was this meant to be just a translation? 
I think you can use weblate for it:
https://hosted.weblate.org/translate/urlcheck/app-strings/en/?checksum=499b92b05d6d461b

Imold:
> Hmm, was this meant to be just a translation? I think you can use weblate for it: https://hosted.weblate.org/translate/urlcheck/app-strings/en/?checksum=499b92b05d6d461b No it was an accident I'm trying to get a fork to build but the workflow gives gives a "Caching for Gradle actions was read-only" error. I accidentally created a pull request while trying to fix it. Honestly I don't know what I'm doing as far as getting it to build this way. 

Also not a  translation just making the title of the quick setting tile shorter so it doesn't become a scrolling title on my phone. "Check URLs from the clipboard" is too long and causes it to over flow but if shortened to "Check URLs in clipboard" it will fit.

PR #397, opened 2024-11-15
Added new resource file system

PabloOQ:
So, I was going to do a Privacy redirector module as per #7. But I noticed all the logic to download files is in `ClearUrlCatalog.java`, to avoid code duplication I got sidetracked working on this.
This is the, 4th(? or 3rd, I don't remember) iteration, but the code is still untested, I don't want to spend more time than necessary in something that might not be merged, althought it _should_ work as it is mainly refactoring.

Here is the class diagram to make things clearer at a glance. I have omitted certain things so it is easier to follow
The main thing to worry is the diagonal of abstract classes as each relies on the previous.
```mermaid

classDiagram
    class ResourceContextInterface{
        <<interface>>
        +getContext() Activity
    }

    class ResourceGuideInterface~T~{
        <<interface>>
        +toObject(String content) T
        +toObjectThrows(String content) T
        +toString(T) String
        +getEmpty() T
    }
    ResourceGuideInterface~T~ <|-- ResourceGuide~T~
    ResourceGuideInterface~T~ <|-- BuiltInResource~T~ : "Delegates to previous"
    ResourceGuideInterface~T~ <|-- ModifiableResource~T~ : "Delegates to previous"
   
    class ResourceGuide~T~{
        <<abstract>>
        +toObject(String content) T
        +toObjectThrows(String content) T*
        +toString(T) String*
        +getEmpty() T*
    }

    ResourceGuide~JSONObject~ <|-- JsonType : JSONObject
    class JsonType{
        + toObjectThrows(String content) JSONObject
        + toString(JSONObject content) String
        + getEmpty() JSONObject
    }

    ResourceGuide~T~ o-- BuiltInResource~T~ 
    class BuiltInResource~T~{
        <<abstract>>
        + getBuiltIn() T
        # buildBuiltIn() T*
    }

    BuiltInResource~T~ <|-- BuiltInResourceFile~T~ 
    class BuiltInResourceFile~T~{
        # buildBuiltIn() T
    }

    BuiltInResource~T~ o-- ModifiableResource~T~
    class ModifiableResource~T~{
        <<abstract>>
        + getCatalog() T
        + clear ()
    }

    ModifiableResource~T~ <|-- ManualEditResource~T~
    class ManualEditResource~T~{
        + save()
    }

    ModifiableResource~T~ <|-- UpdateableResource~T~
    class UpdateableResource~T~{
        + _update(boolean background)
    }
    
    UpdateableResource~T~ <|-- AutoUpdateableResource~T~
    class AutoUpdateableResource~T~{
        + updateIfNecessary()
    }

```
Here is an example to show the actual use.

1. Create a guide to the extension, used for transforming to string/object (this way we can support other fomats like xml)
2. Create a builtin source, either a read only file (from the assets folder) or dynamically built, like the patterns checker module
3. The way this resource will be modified, either manually by the user or it updates from the internet, which can be automatic.

```java
Activity context = null;
// ClearUrls like example
// first, the extension
var type = new JsonGuide();
// second, the builtin source, in this case a file
BuiltInResource<JSONObject> fallback = new BuiltInResourceFile<>(
        context, type, "built-in.json");
// third, how the resource will be modified, in this case it auto downloads
ModifiableResource<JSONObject> modifiable = new AutoUpdateableResource<>(
        fallback, "auto-example.json",
        "example-module",
        "http://fake.domain/catalog",
        "http://fake.domain/hash",
        /*built-in.json-timestamp*/777L/*built-in.json-timestamp*/,
        JavaUtils.weeksToMillis(1));


// Pattern url like example
// first, the extension
type = new JsonGuide();
// second, the builtin source, in this case dynamically built
fallback = new BuiltInResource<>(
        context, type) {
    @Override
    protected JSONObject buildBuiltIn() throws Exception {
        return new JSONObject()

                // built from the translated strings
                .put(context.getString(R.string.mPttrn_ascii), new JSONObject()
                        .put("regex", "[^\\p{ASCII}]")
                )
                .put(context.getString(R.string.mPttrn_http), new JSONObject()
                        .put("regex", "^http://")
                        .put("replacement", "https://")
                )
                ;
    }
};
// third, how the resource will be modified, in this case, manual edit
modifiable = new ManualEditResource<>(
        fallback,"manual-example.json");
```

If you like this change, let me know and I will keep working on the things that will need to be done in order for this to work:
- Change all current implementations of retrieving data from resources, which . Doesn't need to be done all at the same time.
  - JsonCatalog hopefully (only requires changing one variable, hopefully)
    - AutomationRules
    - HostCatalogs
    - PatternCatalog
  - ClearUrlCatalog
- Add an auto migration system (sidetracking again!)
  - Only affects the ClearURLs module, as it stores in the same file both, the updateable catalog and the rules made by the user. The new resource system forces both to be different files.

Let me know what you think! And don't worry if you feel like all this should be scrapped, it is a huge change.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/companions/ResourceCatalogs/AutoUpdateableResource.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/companions/ResourceCatalogs/AutoUpdateableResource.java
@@ -1,0 +1,71 @@
+package com.trianguloy.urlchecker.modules.companions.ResourceCatalogs;
+
+import android.util.Log;
+
+import com.trianguloy.urlchecker.R;
+import com.trianguloy.urlchecker.utilities.generics.GenericPref;
+import com.trianguloy.urlchecker.utilities.methods.JavaUtils;
+
+public class AutoUpdateableResource<T> extends UpdateableResource<T> {
+    protected final int AUTOUPDATE_PERIOD;
+    protected final GenericPref.Bool autoUpdate;
+    protected final GenericPref.Bool lastAuto;
+
+    public AutoUpdateableResource(BuiltInResource<T> builtIn,
+                                  String modifiableFile,
+                                  String key,
+                                  String catalogUrlDefault,
+                                  String hashUrlDefault,
+                                  long lastUpdateDefault,
+                                  int autoUpdatePeriod) {
+        super(builtIn, modifiableFile, key, catalogUrlDefault, hashUrlDefault, lastUpdateDefault);
+        var context = getContext();
+
+        this.autoUpdate = new GenericPref.Bool(key + "_autoUpdate", false, context);
+        this.lastAuto = new GenericPref.Bool(key + "_lastAuto", false, context);
+        this.AUTOUPDATE_PERIOD = autoUpdatePeriod;
+
+        updateIfNecessary();
+    }
+
+    /* ------------------- ui ------------------- */
+
+    @Override
+    public void updateText() {
+        super.updateText();
+        var context = getContext();
+        JavaUtils.ifPresent(txt_check, () ->
+                txt_check.append((lastAuto.get() ? " [" + context.getString(R.string.auto) + "]" : "")));
+    }
+
+    // ------------------- internal -------------------
+
+    @Override
+    public void _update(boolean background) {
+        lastAuto.set(background);   // If called from background, then it must be automatic
+        super._update(background);
+    }
+
+    /**
+     * If the catalog is old, updates it in background. Otherwise does nothing.
+     */
+    private void updateIfNecessary() {
+        var context = getContext();
+        if (autoUpdate.get() && lastUpdate.get() + AUTOUPDATE_PERIOD < System.currentTimeMillis()) {
+            new Thread(() -> {
+                // run
+                lastAuto.set(true);
+                var code = _updateResource();
+
+                // don't show message to user, but log it
+                Log.d("UPDATE", context.getString(code.getStringResource()));
+            }).start();
+        }
+    }
+
+    @Override
+    public void clear() {
+        lastAuto.clear();
+        super.clear();
+    }
+}
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/companions/ResourceCatalogs/BuiltInResource.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/companions/ResourceCatalogs/BuiltInResource.java
@@ -1,0 +1,65 @@
+package com.trianguloy.urlchecker.modules.companions.ResourceCatalogs;
+
+import android.app.Activity;
+
+public abstract class BuiltInResource<T> implements ResourceGuideInterface<T>, ResourceContextInterface {
+    protected final Activity context;
+    protected final ResourceGuide<T> type;
+
+    /**
+     * @param context Can't be null
+     */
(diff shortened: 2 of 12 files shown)


TrianguloY:
Let's see,

It's true I would like to extract the logic for the 'automatic update of a file-like asset', as that will be needed in other modules (even now it will help with the current ones, as you mentioned in the description)

My main concern is that the pr is...big. Not by a lot, I've seen bigger, but still I wonder if all the logic and classes are needed. Maybe they are, but I still need to review everything, and right now I simply can't :(

So, for now I think it's best to leave this. But don't delete it! If I come back to this feature I'll definitely take a look at this and try to adapt it. Thank you for your effort, much appreciated!

Alternatively, maybe what you can try to do (if you want) is to extract the current logic. Just a cut/paste of code, moving things from a file to another, and only adding minimal 'glue logic' when needed. That could help with a later refactor, but since this code is a bit 'entangled' with other things, I don't think it will be easy either...

PabloOQ:
> but still I wonder if all the logic and classes are needed. Maybe they are

All the logic is needed, I think, however, the classes, well that is up for debate. Because the objective was to remove duplicated code, I tried to (strictly) make the responsibilites not interfere with each other, and that led to this. I might have overused the interfaces, and I don't think we will ever need non-json files.

> Alternatively, maybe what you can try to do (if you want) is to extract the current logic. Just a cut/paste of code, moving things from a file to another, and only adding minimal 'glue logic' when needed. That could help with a later refactor, but since this code is a bit 'entangled' with other things, I don't think it will be easy either...

That's the thing, I didn't write much code for this PR. The code for the methods is copied from other places in the source code, and the variables are too, maybe with different names. I avoided refactoring directly so involved classes could work correctly until they were explicitly moved to the new system.

What if I do it like this?
1. Add a substantial amount of comments to each method that directly references the exact origin of that specific code. I was thinking of using the javadoc `@implSpec` annotation, temporarily, until this is finished. This should hopefully make reviewing way easier.
2. Implement the Privacy Redirector Module as an example of use. That way I can also test if everything works correctly, after all, I need a file to test if the file system works. Although that would mean 2 features in 1 PR, which I don't like.
3. No more changes. Because I did not refactor the original classes, but instead copy/pasted, they will keep working correctly. No changes to ANY of these, and no migration either ↓
> - Change all current implementations of retrieving data from resources, which . Doesn't need to be done all at the same time.
>   - JsonCatalog hopefully (only requires changing one variable, hopefully)
>     - AutomationRules
>     - HostCatalogs
>     - PatternCatalog
>   - ClearUrlCatalog
4. Later, when you have the time, or the code is needed, it will be ready for review and merge. You could completely ignore the Privacy Redirector Module and disable or remove it entirely if necessary, to avoid reviewing and even bigger PR. 
5. Remove the comments from the `@implSpec` annotation.
6. If everything works correctly, slowly do what was left behind on point 3.

I honestly don't mind having the PR open, if you don't mind either, and if this is something that at some point in the future will need to be revisited, then I don't see any problem on finishing this now, and leaving it open until review.

If you are ok with this let me know and I will keep working on it.

---

Btw, off-topic, I have a small patch for the automations, it allows for both Strings or arrays of strings in the "regex" field. I felt I was cluttering the file with too many "Unshort X shortener service" and thought it might be useful. I know that technically I can make a regex that supports everything, but this is more accessible and easier to maintain in my opinion. If the decision to have only Strings wasn't made on purpose, do you want me to make a PR on this?

Issue #360, opened 2024-06-02
Option to open a specific Url In seperate browser directly 


heartlog:
### Describe a related problem (optional)

Suppose i want to open https://website.com/* in one browser, url checker don't come that much handy. Add module to pass the url checker and directly open link into my defined browser.

### Describe your suggested feature

If I try to open specific links, it'll open it to my desired browser directly.

### Describe alternatives you've considered for your suggested feature

- Something like default browser for url -

### Acknowledgements

- [X] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I will fill out all of the requested information in this form.

ncoleman:
I support this feature request.  For example, I get emails every day from several vendors.  I don't want to go through URLchecker when I open the links in the email, just go straight to the browser.

I imagine this could work like a pattern match in the hosts fille, or similar.

KaXaSA:
Version 3.0 is almost there with the "Automations" feature. While you might not be able to specify the browser, the app should remember the last browser used for each URL and automatically open it in that browser directly.

Example (Google)

1. Click any link that has the Google domain (ex: [https://google.com](https://google.com))

1. Select the Browser you want it to use, I will use FireFox for this example.

1. Now go back to **URLCheck**

1. Click **Automations** and enable both options 

1. Click **Json editor**
 (probably a good idea to read the instructions)

1. Paste the code below inside it, replacing all the content:
  ("Unshort bit.ly" is the default example that is already in the file, I just added a new entry with my Google example)
```
{
  "Unshort bit.ly": {
    "regex": "https?:\/\/bit\\.ly\/.*",
    "action": "unshort",
    "enabled": false
  },
  "Google": {
    "regex": "https?:\/\/(www\\.)?google\\.com.*",
    "action": "open",
    "enabled": true
  }
}
```


7. Save

8. Click on any link from Google's domain.
 (it should open it directly in the browser you selected in step 2.)

---

I would love if we could actually define the browser/app that will be used to open the matched regex. 

Also, it would be nice to allow an array/list in the regex, like: 

```
    "regex": [
        "https?:\/\/(www\\.)?site1\\.com.*", 
        "https?:\/\/(www\\.)?site2\\.com.*"
    ],
```

I understand you can adapt the regex to archive something similar, but things gets ugly pretty fast.

PabloOQ:
[...]

PR #399, opened 2024-11-22
Allow an array or a value in automations regex field

PabloOQ:
Mentioned in #360 .

In the Automations, the regex field now allows use of both, arrays of Strings or a single String value.
Makes it easier to maintain a big file and more is accessible to add/remove regex to an individual rule.


TrianguloY:
Personally I would have kept the getString method, because that will (according to the documentation) coerce the value (so, even if it's not a string, it will always return a string). This could be done also calling toString anyway.
But I like your generic approach to allow any type, not sure if there will be a need for something similar in the future, but looks good. Let's keep it for now.

I think I would have also created a new JsonUtils class, since these are not _technically_ java utils, but that can be created later if more methods are added, so not a problem.

The only missing thing is to change the auto_editor string to explain this new option, but I just added it myself :)

(also, the code at `com/trianguloy/urlchecker/modules/list/PatternModule.java:156` could be extracted to use this new method, although you will need to convert it to an opt...which I may do in the future, so don't worry about that for now)

PR #423, opened 2025-02-24
IzzyOnDroid modifications

PabloOQ:
[URLCheck is now available on IzzyOnDroid 🎉](https://apt.izzysoft.de/fdroid/index/apk/com.trianguloy.urlchecker) 

- Add IzzyOnDroid badges
- Remove blob DEPENDENCY_INFO_BLOCK

The blob removal was [requested by izzy](https://gitlab.com/IzzyOnDroid/repo/-/issues/719#note_2362817523), more info about it [here](https://android.izzysoft.de/articles/named/iod-scan-apkchecks?lang=en#blobs). From what I've read, it is unnecesary, doesn't do anything and (although very unlikely) could result in potential issues. So it makes sense to remove it.

Since these are small changes I'll merge them myself, hope you don't mind.


TrianguloY:
🥳 thank you!
Happy to see it already available too!

As for the changes, as long as that doesn't change the final apk and users can still update the app without noticing, perfect for me (it's strange because the app doesn't require any external third party library, just the android sdk and the code from this repo, so it's even better to remove any google extra unwanted bits)

Thank you for taking the time to create and manage the integration, much appreciated! 😊
I'll try to add a comment on the izzy issue to thank them too.

PR #441, opened 2025-03-17
Add commonly used libredirect URLs

### LINKED ISSUE #439: Use Android's "Open by default / Verified Links" functionality
### Describe a related problem (optional)

_No response_

### Describe your suggested feature

Coming here as a previous UntrackMe user, this app is missing one huge feature - the ability to make certain links open specifically in this app. As of now, you can set URLCheck to be the "browser app" and open all links in it. The problem with that is, that links inside of a browser do not open in URLCheck. With UntrackMe, what used to be possible is that the browser would think that on device, there was an app to handle the URL, but in reality, it was UntrackMe taking that URL, and opening it back in the browser, but redirected. Basically like this:

- User opens a reddit link in Brave browser
- Brave thinks there's a reddit client installed, URL goes to URLChecker
- URLChecker applies all the necessary rules
- The new URL opens back in the browser

### Describe alternatives you've considered for your suggested feature

_No response_

### Other details

_No response_

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

ArtikusHG:
Fixes #439, adds URLs commonly used for redirection to privacy-respecting front-ends to the manifest, to allow system-wide usage.

Feel free to remove some of them if that's too much.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -73,6 +73,64 @@
                 <data android:mimeType="text/plain" />
                 <data android:mimeType="application/xhtml+xml" />
             </intent-filter>
+
+            <!-- Commonly used libredirect URLs -->
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <!-- Instagram -->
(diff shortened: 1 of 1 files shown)


TrianguloY:
Let's add it for now, if useful I don't mind adding more in the future.

PR #456, opened 2025-03-29
Youtube link cleaner module

Final-Phoenix:
Novice attempt at adding a module for cleaning Youtube links. It works, but please look over my work and test it out as I have not fully tested it. It's might not be complete or could be potentially buggy. Feel free to remove/add anything you see fit (or reject the pull request entirely) as this is my first time trying a pull request on a relatively simple app.

When a user opens a YouTube link, it will clean the URL by removing tracking parameters while keeping the essential parts of the URL intact.

1. Main Features:

- Automatically detects YouTube URLs (i.e. youtube.com, m.youtube.com, youtu.be, and other variations) and removes tracking parameters like:
si (session ID)
feature (feature tracking)
utm_* parameters (campaign tracking)
fbclid (Facebook click ID)
mc_cid and mc_eid (Mailchimp tracking)
_ga (Google Analytics)
gclid (Google Click ID)

2. Configuration Options:

- Enable/Disable: Toggle the main module
Auto-clean Youtube URLs: Automatically clean URLs without manual intervention
Show detailed information: Show detailed information about what was cleaned

3. Status Messages:

- Shows status messages indicating:
When a URL is cleaned
When no changes were needed
When the URL is not a YouTube link
Any errors that occur during cleaning

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -249,10 +249,14 @@ Note: if you edit the patterns, new built-in patterns from app updates will not 
     <string name="mOpen_tabsDesc">Toggle Custom Tabs feature</string>
     <string name="mOpen_incognitoDesc">Toggle incognito mode (for Firefox only)</string>
     <string name="mYoutubeCleaner_name">YouTube Link Cleaner</string>
-    <string name="mYoutubeCleaner_desc">Removes tracking parameters from YouTube links to enhance privacy and reduce tracking.</string>
-    <string name="mYoutubeCleaner_noChange">No tracking parameters found to remove.</string>
-    <string name="mYoutubeCleaner_notYoutube">Not a YouTube URL.</string>
-    <string name="mYoutubeCleaner_error">Error processing URL.</string>
+    <string name="mYoutubeCleaner_desc">Removes tracking parameters from YouTube URLs</string>
+    <string name="mYoutubeCleaner_auto">Auto-clean YouTube URLs</string>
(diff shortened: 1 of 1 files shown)


TrianguloY:
That action should compile the application using gradle, similar to how android studio does when you use the ui. I can try to investigate, don't worry, or try to debug since the action is publicly available.

I see however some changes to files not related to the module, are they something expected?

I also have some doubts about the module itself, as I think the functionality is already performed with the url cleaner module (or can be added manually).

I'll try to do a fast review in the following days, but I'm not known for those :( sorry

Just one question: is this for a school/university assignment? (Not a problem if it is, other students have created prs, just asking to know how fast should I approach this).

TrianguloY:
As you prefer, but please don't let this discourage you! It's true that this module doesn't seem necessary, but I glanced over the code and the structure seems ok at first look. It's true that there seems to be some unnecessary checks, and some of the app custom methods may not be correctly used, but that's on me as I admit my code sometimes relies a lot on custom methods.

I hope at least that this was fun and interesting, and that you learned about android. And please don't hesitate to do something else in the future! As a general recommendation that I always give, try to comment on a related issue or create a new one if you plan to propose changes, as the maintainers may give you tips of point you in the general direction, to avoid wasting your time. You can also look for the 'good first issue' tag that is usually (not always, but I do use it) for issues that should be easy. I also suggest to start with small changes, but in this case a new module without much logic fits into that category so that's ok 👍

PabloOQ:
> please don't let this discourage you!

Agree! I was a student when I did my first PR here. That module is deprecated now 😅, but I still contribute

Issue #493, opened 2025-06-14
Crash when using HTTP URL for webhook


kidhanis:
### Steps to reproduce

1. In Modules, open Webhook
2. In "Webhook URL", add the "http:" prefix. Then, choose any Template.
3. Click on "Test webhook"

### Expected behavior

A message saying "Failed to send webhook"

### Actual behavior

App crashes. Logcat shows 
```java.lang.ClassCastException: com.android.okhttp.internal.huc.HttpURLConnectionImpl cannot be cast to javax.net.ssl.HttpsURLConnection```

### URLCheck version

3.4

### Android version

Android 11

### Android Custom/Specific ROM or Device

Samsung Galaxy S10e

### Other details

In commit bc252d0, I noticed that the webhook module previously called `HttpUtils.performPOSTJSON`, which cast `URLConnection` to `HttpURLConnection`. The Connection wrapper that replaced this method casts to `HttpsURLConnection` instead.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

TrianguloY:
Ugh, your are correct. After the refactor I'm casting everything to https, I missed that. Hopefully it's just a matter of checking and managing both options!

Thanks for the report

kidhanis:
Looking at [Connection.java](https://github.com/TrianguloY/URLCheck/blob/afe102022f96c4c472c201ff354969b6820e1c95/app/src/main/java/com/trianguloy/urlchecker/utilities/wrappers/Connection.java), I see that the [HttpsURLConnection](https://developer.android.com/reference/javax/net/ssl/HttpsURLConnection) methods in use are all inherited. Would casting instead to its superclass `HttpURLConnection` suffice?

kidhanis:
Sure, I'm on it

PR #507, opened 2025-07-23
Replace HttpsURLConnection with HttpURLConnection

kidhanis:
Fix for #493

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/utilities/wrappers/Connection.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/utilities/wrappers/Connection.java
@@ -16,9 +16,9 @@ import java.net.URL;
 import java.net.URLEncoder;
 import java.util.Map;
 
-import javax.net.ssl.HttpsURLConnection;
+import java.net.HttpURLConnection;
 
-/** Wrapper of HttpsURLConnection with some util methods */
+/** Wrapper of HttpURLConnection with some util methods */
 public interface Connection {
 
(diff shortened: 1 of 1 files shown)


TrianguloY:
The action is building the apk from the master code, not yours.

It is wrongly configured on my part (and never noticed). Let me fix it and I'll rerun it.

TrianguloY:
I'm having issues with the action permissions, I may take a bit to fix those.

In the meantime you can download the apk from the [action summary](https://github.com/TrianguloY/URLCheck/actions/runs/16464271104?pr=507)

kidhanis:
On that apk, there's no casting error. Tapping "Test webhook" with `http:` as the "Webhook URL" now proceeds up to a crash after an `ASSERT_ERROR: Unable to post string`. Is this an expected crash for the test app?

TrianguloY:
I'm afraid not.

The test version does have the debugging enabled, and that's why it crashes (on the non-debug version it would simply continue and display a generic "could not send" error).

But the crash itself should not happen, assuming the url is valid and all that.

In order to see the exact exception you can try to tap my icon in the about screen and search there. If you can't find it you will need to launch the app using a debugger.

I'll try to review myself in any case, but I may take a while.

kidhanis:
Sorry, I was testing with `http:` as the URL, the same input I had been using to reproduce the issue. I wasn't testing with a working webhook.
 
When I use a valid HTTP webhook URL, the module sends the payload as expected and the app doesn't crash.

BTW, I can't select any text from the Logcat in the About screen. Is this normal?

PR #476, opened 2025-05-16
Migrate from Triple-T to Fastlane

PabloOQ:
So, apologies for the delay, life happened the weekend I had planned to do this. After that I couln't finish it and now we are here.

Anyways, I moved the folders by hand, I don't know if there are any tools to do this, I didn't find any, and I wasn't feeling like doing a script. Hopefully there are no mistakes, although there are still things to be done, I'll research on how to solve the issues

Also, I know you don't like many commits on the same PR, but this way it is easier to keep track on what files moves to where.

Resources:
- https://gitlab.com/IzzyOnDroid/repo/-/wikis/Fastlane
- https://gitlab.com/-/snippets/1895688

TODO:
- [ ] Look for gradle plugins/integration
- [ ] Screen capture automation
  - Or at the very least, remove the frame from the current images
- [ ] Modify `tools/fixChangelogs.kts`
- [ ] Research if missing files are a problem [(see)](https://github.com/PabloOQ/UrlChecker/actions/runs/15052014093/job/42308888378)
- [ ] Fix `.github/workflows/release.yaml`
- [ ] Fix `app/src/main/java/com/trianguloy/urlchecker/activities/AboutActivity.java`
- [ ] Check if the Changelog module works properly
- [ ] Change [store's metadata file mask](https://hosted.weblate.org/projects/urlcheck/play-store/#information) to `fastlane/metadata/android/*` (and default to `fastlane/metadata/android/en-US`)

PR #513, opened 2025-08-19
fastlane from IzzyOnDroid & integrating metadata from Triple-T

IzzySoft:
this PR provides you with a "fastlane starter package" (as discussed in #476) – to give it into your hands to define how your app is presented, and make it easy to keep descriptions & graphics in sync with your development. Some notes to sum them up in a place easy to find for you:

* the [IzzyOnDroid Fastlane Documentation](https://gitlab.com/IzzyOnDroid/repo/-/wikis/Fastlane) can guide you with maintaining the structures here
* take special care for limits, e.g. per-release changelogs can have max 500 chars, `full_description.txt` max 4,000 chars, graphics must have specific aspect ratios etc.
* you can have "bigger graphics" here; the IoD updater takes care to resize/optimize them as needed
* for the `full_description.txt` here I've used HTML compressed into a single line, to prevent the fdroid software converting each line break to a `<br>`. The tags used for this are supported by F-Droid.org as well, and to my knowledge even by PlayStore.
* whenever you add a new "type" of metadata that has not been there before (e.g. the first changelog, a featureGraphic if it wasn't yet present), please let us know so we can check and enable it.
* Details on all this in the linked documentation.
* Once merged, your changes will be synced whenever a new release is pulled in here – to make sure it stays up-to-date with the app itself.
* Should you have questions, be welcome to ask :wink:

And now: Enjoy!

### DIFF
--- a/fastlane/migration.md
+++ b/fastlane/migration.md
@@ -1,0 +1,7 @@
+# TripleT -> Fastlane migration
+
+This folder contains all the Fastlane-related files copied from the Triple-T ones (by Izzy).
+Check https://github.com/TrianguloY/URLCheck/pull/513 for some more details.
+
+A proper TripleT -> Fastlane migration is still pending.
+Check https://github.com/TrianguloY/URLCheck/pull/476 and https://github.com/TrianguloY/URLCheck/issues/425 for the current status.


IzzySoft:
Yeeeha! Thanks! :hugs: Just enabled it on our end and manually triggered a sync. Now all those locales are present here as well (before, we just had en+de) – and will show up in our catalog with the next sync around 6 pm UTC. Looked great to see the import flow through!

Thanks also for the documentation. Should any question toward the structure come up, feel free to give me a ping and I'll try to help. Cannot help much with the workflow integration, though, as that's not exactly in the "core" of my area :see_no_evil: 

> I wonder if now the other pr will have less files...

It should. And probably show some conflicts, too – and definitely need a rebase. But I guess all you need from there are the action parts – validation actions and such. Basically, what is in the check list of the other PR's initial post (eh, nice to see it even links to our documentation). Oh, and the adjustments to the Readme (before you finally remove the files from Triple-T).

Not sure where the per-release changelogs in that other PR came from – they should be taken along as well. Not sure what approach is best there, cherry-picking, or rather a fresh and clean start-over…

PR #523, opened 2025-10-01
Add deny browser to open/share functionality

NicholasThyssen:
I saw that someone had requested for there to be less browsers based on user preference so I was able to add something that removed one browser from chrome, edge, opera, brave and Firefox that user did not like. In future, could update to include more apps and maybe even increase the amount of apps available to deny at a single time to be an actual list.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/companions/DenylistOption.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/companions/DenylistOption.java
@@ -1,0 +1,50 @@
+package com.trianguloy.urlchecker.modules.companions;
+
+import com.trianguloy.urlchecker.R;
+import com.trianguloy.urlchecker.utilities.Enums;
+
+/**
+ * Enum representing the app that can be hidden from "Open with" and "Share via".
+
+ * Each entry has:
+ * An id (for saving as a preference),
+ * A label (string resource for UI),
+ * The actual package name of the app to filter out.
+
+ */
+
+public enum DenylistOption implements Enums.IdEnum, Enums.StringEnum {
+    NONE(0, R.string.denylist_none, null),
+    CHROME(1, R.string.denylist_chrome, "com.android.chrome"),
+    FIREFOX(2, R.string.denylist_firefox, "org.mozilla.firefox"),
+    EDGE(3, R.string.denylist_edge, "com.microsoft.emmx"),
+    OPERA(4, R.string.denylist_opera, "com.opera.browser"),
+    BRAVE(5, R.string.denylist_brave, "com.brave.browser");
+
+    private final int id;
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/trianguloy/urlchecker/modules/companions/ShareUtility.java
```diff
@@ -75,19 +83,52 @@ public void onInitialize(View views) {
 
         /** Shares the url as text */
         public void shareUrl() {
```
TrianguloY, line 85: I understand this change, but for now I prefer to leave the share as just 'call android', without extra logic. It may change in the future though, but it will require other functionality present in the open that is still not here (like reordering)

app/src/main/res/layout/config_open.xml
```diff
@@ -119,4 +120,22 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="match_parent"
+        android:orientation="horizontal">
+
+        <TextView
+            android:id="@+id/textView5"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:gravity="center_vertical"
+            android:text="Deny Browser" />
```
TrianguloY, line 133: move this text to the translations file (place it under mOpen_iconSize)

app/src/main/java/com/trianguloy/urlchecker/modules/companions/DenylistOption.java
```diff
@@ -0,0 +1,50 @@
...
+ * Enum representing the app that can be hidden from "Open with" and "Share via".
+
+ * Each entry has:
+ * An id (for saving as a preference),
+ * A label (string resource for UI),
+ * The actual package name of the app to filter out.
+
+ */
+
+public enum DenylistOption implements Enums.IdEnum, Enums.StringEnum {
```
TrianguloY, line 16: Even though this is a good first implementation, I wonder if it's too restrictive.
You can only choose from a predefined list, and you can't choose more than one.

I prefer to store a multilinestring, where each line is one package (empty or invalid lines ignored). This will either require implement a chooser dialog dynamically loaded with the apps...or just leave the textarea and let the user write the packages (not user friendly, I know, but more powerful).

Or, even better, with something similar to the 'templates' in the webhook module. Yes, I think this will be easy to implement, and could be easily replaced with a more powerful dialog if needed.

TrianguloY (changes requested):
Hi, thanks for your contribution!
As you have probably noticed, I'm currently a bit absent, so apologize in advance if it takes me several days to reply. Feel free to ping me though (but don't abuse it, please).

I wrote some comments about the new functionality, try to follow them, but feel free to discuss if you think otherwise.

PR #567, opened 2026-04-10
Restrict songlink pattern to music streaming URLs

### LINKED ISSUE #566: Songlink replacement on every URL
### Steps to reproduce

Tap on any unrelated URL.

### Expected behavior

Songlink replacement should not be shown

### Actual behavior

Songlink replacement option is shown

### URLCheck version

3.5

### Android version

16

### Android Custom/Specific ROM or Device

Doesn't matter

### Other details

The regex is broken, it runs for any link.

### Acknowledgements

- [x] I have searched the existing issues and this is a new ticket, **NOT** a duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I will fill out all of the requested information in this form.

utafrali:
Fixes #566

The songlink module was replacing every URL. Added a regex pattern to only match music streaming services so the replacement doesn't interfere with other links.

### DIFF
--- a/app/src/main/java/com/trianguloy/urlchecker/modules/companions/PatternCatalog.java
+++ b/app/src/main/java/com/trianguloy/urlchecker/modules/companions/PatternCatalog.java
@@ -84,7 +84,7 @@ public class PatternCatalog extends JsonCatalog {
                         .put("enabled", false)
                 )
 
-                // excludeRegex example
+                // excludeRegex example. Consider replacing with another service that allows all urls
                 .put("URL ➔ Songlink", new JSONObject()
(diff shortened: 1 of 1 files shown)


TrianguloY:
Honestly I would like to use a different example, and this is disabled by default so it's not really a big issue, but thanks for the pr. Approved!

siddhpant:
> this is disabled by default

It was enabled by default, I had to manually disable it.

TrianguloY:
> > this is disabled by default
> 
> It was enabled by default, I had to manually disable it.

That's odd, the default entry has the "disabled: false" property 🤔

Are all the other default patterns enabled too?

siddhpant:
I don't think so. Though it has been a long time :P

Maybe something happens when a new rule is added in an update?

